PR #36, opened 2018-10-09
ES6 example  

olzzon:
Simple Javascript ES6 example for copy-pasting, so you can get started quickly.


olzzon:
Had some baldewit in my PR, so I cancel this one

PR #67, opened 2019-02-22
chore: update dependencies

Julusian:

Update dependencies to pass the yarn audit


Julusian:
@nytamin This has broken the tests as the ts compiler/linter is being more strict about the tests and mocks.

Before I or anyone fixes that all up, is this desired or is that something we want to suppress?

Some examples of the errors (its a very long list but I can only see a couple of types):
```
jest.diagnostics` option):
    src/devices/__tests__/abstract.spec.ts:90:41 - error TS2493: Tuple type '[]' of length '0' has no element at index '1'.

    90  	expect(commandReceiver0.mock.calls[0][1]).toMatchObject(


    src/__mocks__/atem-connection.ts:123:41 - error TS6133: 'key' is declared but its value is never read.

    123  setDownstreamKeyOnAir (onAir: boolean, key?: number): Promise<any> {
        
    src/devices/__tests__/casparcg.spec.ts:473:10 - error TS2532: Object is possibly 'undefined'.

    473  	expect(commandReceiver0.mock.calls[4][1].name).toEqual('ScheduleSetCommand')
         
```

PR #74, opened 2019-03-20
fix: Some callbacks not being sent

Julusian:

Fixes some callbacks not being sent

Just before this `_.each` is a `this._doOnTime.clearQueueNowAndAfter(tlState.time)`.
If any of the callbacks get removed by that, then their `sentCallbacksOld[callbackId]` will still report a value, causing them to never be sent. 

This can happen when there are multiple state changes close to each other (I assume), so that the callbacks from the first have not been sent by the time the second is being processed.

If the callback will have been removed by the call to `clearQueueNowAndAfter` then ignore `sentCallbacksOld`

* **Other information**:

More common with multithreaded resolving. Likely due to the state being processed more on time than before (where it was often done late)


Julusian:
Replaced by #75

PR #87, opened 2019-07-30
WIP: Caspar AB playback

Julusian:
This works well enough to develop caspar based AB playback in blueprints, but has some major limitations blocking real usage.

When the layer is clear allow doing a LOAD of the media instead of LOADBG, so that the first frame is visible to the user.

### TODO
- [x] Control behaviour from option
- [x] When `LOAD` fails, previous clip is still playing (need to do a `LOAD empty` first?)
- [x] Be able to do a `LOAD` without parameters to promote background clip to foreground (Needs support in ccg first)
- [x] Same clip back to back on a layer doesn't reload (due to `playing: false, startTime: null` which are both needed to pause on the first frame)
- [ ] Caspar support notOnAir property

### DIFF
--- a/src/devices/casparCG.ts
+++ b/src/devices/casparCG.ts
@@ -457,7 +457,10 @@ export class CasparCGDevice extends DeviceWithState<TimelineState> {
 
 				if (!layerExt.isLookahead) { // foreground layer
 					const prev = channel.layers[mapping.layer] || {}
-					channel.layers[mapping.layer] = _.extend(stateLayer, _.pick(prev, 'nextUp'))
+					channel.layers[mapping.layer] = {
+						...stateLayer,
+						nextUp: prev.nextUp
+					}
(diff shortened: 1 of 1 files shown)


nytamin:
@Julusian what's the status on this one?
I'm inclined to close this PR due to inactivity, to be reopened when there is a need and it's ready for merging.

Julusian:
I think that at least some of this will be needed for tv2, so does need to be done soon.
Im not aware of that being tracked in asana though

thomasslee97:
@nytamin This is where I pulled the AB code from that's making TSR tests fail on TV 2s fork

Julusian:
This is replaced by https://github.com/olzzon/tv-automation-state-timeline-resolver/tree/feat/ab-playback
A PR shall be raised for that once it has been sufficiently tested

PR #100, opened 2019-08-27
fix: hyperdeck: Only emit connectionChanged event when status actually changes

nytamin:
(Disclaimer: I actually haven't test this code, only done static analysis)

I believe that the connectionChanged event is emitted continously from the hyperdeck device.

This fix aims to make in only emit when the status has actually changed


baltedewit:
I'd prefer not to have to do a deep object comparison every time we check recording time (down to 10s intervals) so I've commited a different fix for this: https://github.com/nrkno/tv-automation-state-timeline-resolver/commit/7f1f6de99d6da810f1335dc31a01a541c342d079

PR #118, opened 2019-10-11
feat: use a timeline obj to retrigger lawo commands

baltedewit:
This PR removes the use of the triggerValue on the lawo timeline objects and instead looks for an object on the timeline with a triggerValue. This means that timings for retriggering can be more precise.

In addition, removing a triggerValue from the timeline will no longer retrigger commands as we consider retrigger to be an action and not a state.

### DIFF
--- a/src/devices/__tests__/lawo.spec.ts
+++ b/src/devices/__tests__/lawo.spec.ts
@@ -32,8 +32,14 @@ describe('Lawo', () => {
 			mappingType: MappingLawoType.SOURCE,
 			identifier: 'BASE'
 		}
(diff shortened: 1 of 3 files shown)


Julusian:
@baltedewit Are the layers being sorted before being applied to `lawoState`?
Unlikely to be a problem and not directly related to this change, but would be nice to have in general

Julusian:
@baltedewit also, prettier is complaining about formatting

baltedewit:
> Are the layers being sorted before being applied to `lawoState`

No, later on the commands are sorted on priority though.

PR #120, opened 2019-10-15
fix(atem): audio channel tsr objects are correctly merged

baltedewit:
Previous behaviour:

Audio channels would not be controlled by TSR at all

New behaviour:
Timeline objects are merged correctly into the state objects and TSR controls the audio channels on the atem


Julusian:
@baltedewit how was this not picked up as a typing issue?

PR #124, opened 2019-11-06
fix: estimateResolveTime was unused due to avoiding any resolves for times in the past

Julusian:
I cant find anything in Kibana that was using the estimateResolveTime codepath.

Reading through this confirms that it could never be reached too (unless now was somehow 0). This should work the same, but factor in estimatedResolveTime


nytamin:
Terje and Jan T reports that this fix seems to work in the field.

Julusian:
@nytamin It would be nice to check the timings in the logs, if possible

PR #128, opened 2019-11-25
fix: bug fix: fps

nytamin:
I encountered an issue today which seems to come from the channel.fps having been set inverse.

It seems like we should use fps and not frame-time.

Ref: The fps is used (among other places) here:
https://github.com/SuperFlyTV/casparcg-state/blob/develop/src/lib/transitionObject.ts#L110

TODO: To be discussed


baltedewit:
The calculations in your reference example assume frame-time, but the default is in fps. Looks like the library needs to be consistent before we can do anything in TSR.

nytamin:
Closing this as semi-invalid.
I have started on a project in ccg-state to make typings stronger and clearer what is frames and what is milliseconds, but the reason for this PR is invalid, as I has mixed up seconds and milliseconds...

PR #129, opened 2019-12-05
Fix/node12 multithreaded devices

Julusian:
Uses latest threadedClass, and resolves errors when running with node12

This is not yet complete. While playout works, there are still some data clone errors/uncaught exceptions logged.

Attempting to run with node12 crashes on startup unless multi-threaded devices is disabled

* **Other information**:
The primary issue is that threadedClass would never be able to pass in `getCurrentTime` to the child due to it silently failing. This became a fatal error in node 12 and so was noticed.
Part of the problem was that the function was inside an object. This has had to be flattened out to get it to work. However this means that it gets mangled which causes problems for jest (hence some of the commented hacks)


Julusian:
@nytamin I am not noticing any more errors.
I think later on we should rethink some of the TSR device api, once we have worked on strictening up the threadedclass api and figuring out to minimise and easily test for data clone errors

Julusian:
This needs to be changed to bump the version number more than a patch release.

PR #132, opened 2019-12-18
feat: atem media pool config

jstarpl:

This PR introduces an interface for defining assets to be uploaded to the ATEM

The setting used to be a property on the Studio, but this is gone now and there is no way to set up what files need to be uploaded to the ATEM.

There is a new property on the AtemOptions called mediaPoolAssets

* **Other information**:


Julusian:
This probably doesnt want to be merging the version number change into develop?
Other than that its good.

PR #137, opened 2020-03-26
Feature/smooth casparcg transitions

nytamin:
This PR adds the ability to let TSR run CasparCG transitions internally, by iterating the transitions and sending many update commands instead.

This works by intercepting MIXER commands just before they are sent, and instead using their payload to start iterating the internal transitions

Note: This PR is dependent on https://github.com/SuperFlyTV/casparcg-state/pull/53

TODO:

- [x] Support for FILL
- [x] Support for PERSPECTIVE
- [ ] Support for combining others with this

Example timeline object (tested with quickTSR):

```
import { DeviceType, TimelineContentTypeCasparCg, Transition, TSRTimeline, Ease } from 'timeline-state-resolver'
import { TSRInput } from '../src'
import { literal } from 'timeline-state-resolver/dist/devices/device'


export const input: TSRInput = {
	timeline: [
		{
			id: 'video0',
			enable: {
				start: Date.now(),
				duration: 20 * 1000
			},
			layer: 'casparLayer0',
			content: {
				deviceType: DeviceType.CASPARCG,
				type: TimelineContentTypeCasparCg.MEDIA,
				file: 'amb',
				mixer: {
					fill: {
						x: 0.0,
						y: 0.0,
						xScale: 0.25,
						yScale: 0.25
					}
				}
			},
			keyframes: [
				{
					id: 'kf0',
					enable: {
						start: 500
					},
					content: {
						mixer: {
							fill: {
								x: 0,
								y: 0.75,
								xScale: 0.25,
								yScale: 0.25
							},
							changeTransition: {
								type: Transition.INTERNAL,
								duration: 0, // not used
								easing: Ease.INTERNAL_PHYSICAL,
								direction: hackDirection
							}
						}
					}
				},
				{
					id: 'kf1',
					enable: {
						start: '#kf0 + 1000'
					},
					content: {
						mixer: {
							fill: {
								x: 0.45,
								y: 0.45,
								xScale: 0.25,
								yScale: 0.25
							},
							changeTransition: {
								type: Transition.INTERNAL,
								duration: 0, // not used
								easing: Ease.INTERNAL_PHYSICAL,
								direction: hackDirection
							}
						}
					}
				}
			]
		}
	]
}
```


nytamin:
There, I'm pretty happy with this now and ready to accept reviews on it.

src/types/src/casparcg.ts
```diff
@@ -429,21 +433,38 @@ export interface Mixer {
...
 	easing: Ease
-	direction: Direction | string
+	direction: Direction
+}
+export interface TSRTansition {
+	type: Transition.TSR_TRANSITION
+	customOptions: TSRTransitionOptions
 }
+
+export type Transition0 = CasparCGTransition | TSRTansition
```
baltedewit, line 470: Transition0 is not a very descriptive name for this

src/devices/transitions/transitionHandler.ts
```diff
@@ -0,0 +1,178 @@
...
+		options: TSRTransitionOptions,
+		animatorTypes: {[groupId: string]:
+			{
+				type: 'linear' | 'physical',
+				options?: TSRTransitionOptions
+			}
+		},
+		updateCallback: (newValues: number[]) => void
+	) {
+		// Note: this is a preliminary implemenation, that animates x & y, and snaps xScale & yScale
```
baltedewit, line 84: Which bit is preliminary? It seems to iterate generically. But if this is still true it would be better to put it into a jsdoc style comment so the editor will show it when using the method.

src/devices/transitions/transitionHandler.ts
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	public clearTransition (identifier: string) {
+		const t = this._transitions[identifier]
+		if (t) {
+			this._stopTransition(t)
+			delete this._transitions[identifier]
+		}
+	}
+	public snapTransition (
```
baltedewit, line 40: Would this method name be more self-explanatory if it was called `stopAndSnapTransition`?

src/devices/casparCG.ts
```diff
@@ -897,4 +909,203 @@ export class CasparCGDevice extends DeviceWithState<State> implements IDevice {
...
+					const opt: { initial: number, prop: string } = (
+						objectParams.keyword === 'OPACITY' ?
+						{
+							initial: 1,
+							prop: 'opacity'
+						} :
+						objectParams.keyword === 'VOLUME' ?
+						{
+							initial: 1,
+							prop: 'opacity'
```
baltedewit, line 1037: Given the keyword I would assume this should be `volume`?

baltedewit (changes requested):
Implementation looks pretty good, but I would expect additional functionality to include new unit tests for said functionality.

src/devices/casparCG.ts
nytamin: Congratulations! you found my test-if-the-reviewer-reads-all-of-the-code bit ;)

src/devices/__tests__/casparcg.spec.ts
```diff
@@ -1352,7 +1352,7 @@ describe('CasparCG', () => {
 					inputType: 'decklink',
 					deviceFormat: ChannelFormat.HD_720P5000,
 
-					format: ChannelFormat.PAL,
+					// format: ChannelFormat.PAL,
```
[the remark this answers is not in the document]
nytamin: Good question. The `format` attribute is non-existent in https://github.com/nrkno/tv-automation-state-timeline-resolver/blob/feature/smooth-casparcg-transitions/src/types/src/casparcg.ts#L140 I _think_ it's replaced by `deviceFormat` (?)

src/devices/transitions/transitionHandler.ts
nytamin: It was preliminary at the time of writing, should be definite by now 👍

src/devices/__tests__/casparcg.spec.ts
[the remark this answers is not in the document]
baltedewit: Let's just delete the line then, no need to keep old types around :)

baltedewit (changes requested):
The tests are not passing (linting error?) and I don't see any unit tests for the special cases CasparCG._interceptCommand (only regular commands are tested)

nytamin:
@baltedewit I was obviously a bit too trigger-happy there... I've now added a rudimentary test for the functionality in the casparcg-device.

The test is definitely not covering all of the functionality, but I'm inclined to not spend more time on this at the moment and merge as-is (if you agree).
When/If we decide to actually start using the feature again we could spend some more time on adding tests for it then.

baltedewit (approved):
I agree that it should be OK to merge in the current state.

PR #142, opened 2020-05-05
Fix/http makeready can reset state

Julusian:

* **Other information**:


Julusian:
@nytamin as we discussed, there is now a queue in conductor. Instead of using syncFunction, I opted to use an off the shelf queue for promises `p-queue`. I did look at using syncFunction, but that would have required a fair bit of time to convert it from fibers to promises and this was easier.
I have handled the 'ignore' functionality, by doing the `queue.add` after the `_resolveTimelineRunning` guard.
And timeouts for makeReady is handled with the help of `p-timeout`

Additionally, I went though any uses of Promise.all and they are now using `p-all` so that it can block until all the child promises have completed and not reject early

Julusian:
During this I hit what appears to be https://github.com/nrkno/tv-automation-state-timeline-resolver/pull/139, so have left one use of Promise.all which would be nice to remove also

PR #145, opened 2020-06-03
feat: npm hotfix workflow

Julusian:
#### Prerelease (unchanged)
All commits to master will trigger a prerelease (except [ci skip] commits)

Commits made to branches with [publish] will trigger an 'experimental' prerelease

#### New version
Locally run `yarn release` to invoke standard-version, and then push the resulting commit and the newly created tag

#### Hotfix
If you need to do a release between 1.2.3 and an already existing 1.2.4, then that will need to be done as a prerelease, as semver doesnt have any more granular numbering that we can use

This should be done on a hotfix branch, eg `hotfix/1.2.x`
1) Figure out what the correct version number should be. It should be simply a patch version, but if that is already in use that could get more complicated
1) `yarn release --release-as 1.2.3` (substitute in the number it should be)
1) push the resulting commit and newly created tag
1) Ensure the version number in master wont generate the same number on its next release (by setting it to 1.2.4-0 if you just published 1.2.3)

## TODO
* [ ] Restore the `chore: [release]` workflow for triggering a release
* [ ] Branches should only prerelease when they have the [publish] trigger


nytamin:
status on this: julian is happy, but wants to do some finalizing and cleanup before merging this

Julusian:
This is pretty stale now, and will need to be rewritten to github actions before it should be merged

PR #146, opened 2020-06-04
feat: use p-all instead of Promise.all for better handling when one rejects

Julusian:
When using Promise.all, it will reject when the first child rejects. The other children will keep executing until completion.
This will lead to race conditions as we run many of these operations behind a lock to ensure only one happens at a time.

By using p-all, we can tell it to reject only once they have all completed, and so we will retain the lock until they are all done.


nytamin:
Very good point! 👍

PR #149, opened 2020-06-09
ATEM interface documentation

thomasslee97:
- Re-enables the `programInput` property of the `me` object, allowing program to be cut directly. This in turn allows the preview source to be set.
- Adds documentation to the ATEM type definitions, including acceptable value ranges for numeric properties.

### DIFF
--- a/src/types/src/atem.ts
+++ b/src/types/src/atem.ts
@@ -132,10 +132,13 @@ export interface TimelineObjAtemME extends TimelineObjAtemBase {
 			input?: number,
 			transition?: AtemTransitionStyle,
 
-			/** To change previewInput you must use programInput to set program directly */
 			/** Cut directly to program */
 			programInput?: number;
-			/** Set preview input (cut) */
+			/**
+			 * Set preview input.
+			 * Cannot be used in conjunction with `input`;
+			 * `programInput` must be used instead if control of program and preview are both needed.
+			 */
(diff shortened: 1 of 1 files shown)


Julusian:
I could be really picky here and say that some of those ranges are technically wrong. But only if working in a 4:3 videomode :wink: 

More seriously, I am happy with this

src/types/src/atem.ts
```diff
@@ -114,32 +128,49 @@ export interface TimelineObjAtemME extends TimelineObjAtemBase {
 		deviceType: DeviceType.ATEM
 		type: TimelineContentTypeAtem.ME
 		me: {
+			/** Must be used with transition property, sets input to transition to */
 			input?: number,
 			transition?: AtemTransitionStyle,
 
-			// programInput?: number; // programInput exists, bu I don't think we should use it /Nyman
+			/** To change previewInput you must use programInput to set program directly */
```
baltedewit, line 135: Using previewInput implies you can only change the program via programInput and not via input + transititon. In your current wording it seems like you can only change previewInput if you defined programInput which is not true. PR looks good otherwise.

thomasslee97: Good point, updated wording: https://github.com/olzzon/tv-automation-state-timeline-resolver/commit/e4b819e09d8a7bca89b0c6ac8aabe2190427d86d

PR #151, opened 2020-06-18
Feat/audio interface combined faders

nytamin:
This PR gives the timeline author the ability to define multiple Sisyfos-channels in a single object.
This will help with reducing the number of mapping-layers and timeline-objects needed to do basic stuff.

Example:
```
{
  id: 'baseline',
  enable: {
    while: 1
  },
  layer: 'sisyfos_channels_base',
  content: {
    deviceType: DeviceType.SISYFOS,
    type: TimelineContentTypeSisyfos.CHANNELS,

    channels: [
      {
        channel: 1,
        faderLevel: 0.1,
        isPgm: 0
      },
      {
        channel: 2,
        faderLevel: 0.2,
        isPgm: 0
      }
    ],
    overridePriority: -999
  }
},
```
This is an added feature, I've taken measures to ensure backwards compatibility with existing timeline-objects.

### DIFF
--- a/src/devices/sisyfos.ts
+++ b/src/devices/sisyfos.ts
@@ -13,12 +13,16 @@ import {
 import { DoOnTime, SendMode } from '../doOnTime'
 
 import {
-	TimelineState, ResolvedTimelineObjectInstance
+	TimelineState,
+	ResolvedTimelineObjectInstance
 } from 'superfly-timeline'
 import {
 	SisyfosOptions,
-	TimelineObjSisyfosMessage,
-	MappingSisyfos
+	MappingSisyfos,
+	MappingSisyfosType,
+	TimelineObjSisyfosAny,
+	TimelineContentTypeSisyfos,
+	SisyfosChannelOptions
 } from '../types/src/sisyfos'
 import {
 	SisyfosApi,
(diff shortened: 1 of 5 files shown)


src/devices/sisyfos.ts
```diff
@@ -206,18 +223,24 @@ export class SisyfosMessageDevice extends DeviceWithState<SisyfosState> implemen
...
-					visible: true
+					...this.getDefaultStatChannel()
 				}
 			}
 
 			deviceState.channels[ch] = channel
 		}
 		return deviceState
 	}
+	getDefaultStatChannel (): SisyfosChannel {
```
baltedewit, line 234: typo?

src/devices/sisyfos.ts
```diff
@@ -237,30 +263,78 @@ export class SisyfosMessageDevice extends DeviceWithState<SisyfosState> implemen
...
+						channel: foundMapping.channel,
+						overridePriority: content.overridePriority || 0,
+						isLookahead: layer.isLookahead || false,
+						tlObjId: layer.id
+					})
+				} else if (
+					foundMapping.mappingType === MappingSisyfosType.CHANNELS &&
+					content.type === TimelineContentTypeSisyfos.CHANNELS
+				) {
+					_.each(content.channels, channel => {
```
baltedewit, line 297: Still using underscore? :) `Object.entries(content.channels).forEach([i, channel] => { ... })` achieves the same thing

Julusian: Or `for (const [i, channel] of Object.entries(content.channels)) {` to get rid of the extra layer of function and potential context switch

nytamin: true, let's keep that in mind for when we decide to ditch it ;)

Julusian: Yeah, but it would be easier to gradually replace uses whenever we work nearby. Besides, we may not be fully remove it, as there are some useful utilities

PR #153, opened 2020-06-24
Feat: audio interface combined faders for Lawo

baltedewit:
This PR gives the timeline author the ability to define the fader value for multiple Lawo-sources in a single object

Similar to https://github.com/nrkno/tv-automation-state-timeline-resolver/pull/151

In addition it removes the `Fader/Motor dB Value` object in favour of a flat structure

Example timeline object for composites:
```ts
{
	id: 'obj0',
	enable: {
		while: 1
	},
	layer: 'lawo_base',
	content: {
		deviceType: DeviceType.LAWO,
		type: TimelineContentTypeLawo.SOURCES,

		sources: [{
			mappingName: 'lawo_c1_fader',
			value: -6
		},{
			mappingName: 'lawo_c2_fader',
			value: -6
		}]
	}
}
```

New timeline object content structure
```ts
{
	deviceType: DeviceType.LAWO,
	type: TimelineContentTypeLawo.SOURCE,

	faderValue: -6,
	transitionDuration: 750
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,9 +183,7 @@ Pull up a fader, and leave it there
 		deviceType: DeviceType.LAWO,
 		type: TimelineContentTypeLawo.SOURCE,
 
-		'Fader/Motor dB Value': {
-			value: 0
-		}
+		faderValue: 0
 	}
 }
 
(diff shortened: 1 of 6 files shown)


Julusian:
I think it would be better to call `value` `faderLevel` instead. Looking at that interface from the outside, it isnt immediately obvious what `value` does. Also it leaves us scope for adding more properties later if we want to

PR #154, opened 2020-06-30
emit resolveDone 

nytamin:
This PR makes the conductor emit an event at the end of resolving (used for reporting of latency downstream)


nytamin:
merged into #158

PR #156, opened 2020-07-09
Feat/shotoku

baltedewit:
This PR adds support for the Shotoku camera robotics system through the TR-XT ECI interface protocol.

Please note this has not been verified with the actual system as of yet. Besides that the documentation regarding the checksum byte is not entirely clear, so this may well not work at all.


baltedewit:
@Julusian could you check whether the interfaces for the timeline objects look sane to you?

PR #161, opened 2020-09-01
feat: Add External Elements status reporting to Viz device

ianshade:
This PR adds reporting of loading status of External Elements (Pilot) in Viz device, by creating/removing Media Objects in core's collection. 

No way to report if an Element was loaded or not.

New `updateMediaObject` and `clearMediaObjects` events, that have to be handled by the Playout Gateway, are added.
A Media Object is created for every External Elements that is loaded. 
Media objects created by this device have their `collectionId` set to the Id of the device. This collection is cleared on activation and deactivation.

* **Other information**:


nytamin:
I lost this one in the deep deep depths of the inbox.
As previously discussed some time ago, this will be replaced by the upcoming ExpectedPackages feature.

PR #167, opened 2020-09-29
chore: update threaded class to version 0.8.0

jstarpl:

Updates the threadedClass library to latest version, 0.8.0

TSR uses version 0.7.0

* **Other information**:


nytamin:
I looked into the issues with the tests, and I think I've found a good solution. I also fixed a few typing issues that popped up along the way

baltedewit (approved):

PR #168, opened 2020-10-19
feat: set casparcg fps in device options

baltedewit:
This PR adds support to set the fps used in CasparCG State diffing mechanisms and time calculations.

### DIFF
--- a/src/devices/__tests__/casparcg.spec.ts
+++ b/src/devices/__tests__/casparcg.spec.ts
@@ -12,6 +12,7 @@ import {
 } from '../../types/src'
 import { MockTime } from '../../__tests__/mockTime'
 import { getMockCall } from '../../__tests__/lib'
+import { AMCP } from 'casparcg-connection'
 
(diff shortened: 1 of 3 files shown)


src/devices/__tests__/casparcg.spec.ts
```diff
@@ -187,6 +188,89 @@ describe('CasparCG', () => {
 			seek: 25 * 10
 		})
 	})
+	test('CasparCG: Play AMB for 60s in 50fps', async () => {
+
+		const commandReceiver0: any = jest.fn(() => {
+			return Promise.resolve()
+		})
+		let myLayerMapping0: MappingCasparCG = {
```
gundelsby, line 196: Shouldn't all these be consts?

baltedewit: Yeap, resolved this in https://github.com/nrkno/tv-automation-state-timeline-resolver/pull/169

PR #169, opened 2020-11-03
Feat: casparcg ffmpeg filters

baltedewit:
This PR introduces video and audio filters to the CasparCG implemenation for use with CasparCG 2.3.

_Note that this PR is a based on https://github.com/nrkno/tv-automation-state-timeline-resolver/pull/168, which has its own review process_

### DIFF
--- a/src/devices/__tests__/casparcg.spec.ts
+++ b/src/devices/__tests__/casparcg.spec.ts
@@ -35,17 +35,17 @@ describe('CasparCG', () => {
 		const commandReceiver0: any = jest.fn(() => {
 			return Promise.resolve()
 		})
-		let myLayerMapping0: MappingCasparCG = {
+		const myLayerMapping0: MappingCasparCG = {
 			device: DeviceType.CASPARCG,
 			deviceId: 'myCCG',
 			channel: 2,
 			layer: 42
 		}
-		let myLayerMapping: Mappings = {
+		const myLayerMapping: Mappings = {
 			'myLayer0': myLayerMapping0
 		}
 
-		let myConductor = new Conductor({
+		const myConductor = new Conductor({
 			initializeAsClear: true,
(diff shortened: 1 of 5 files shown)


src/devices/__tests__/casparcg.spec.ts
```diff
@@ -187,6 +188,89 @@ describe('CasparCG', () => {
 			seek: 25 * 10
 		})
 	})
+	test('CasparCG: Play AMB for 60s in 50fps', async () => {
+
+		const commandReceiver0: any = jest.fn(() => {
+			return Promise.resolve()
+		})
+		let myLayerMapping0: MappingCasparCG = {
```
gundelsby, line 196: Shouldn't all these be consts?

src/devices/casparCG.ts
```diff
@@ -467,11 +480,11 @@ export class CasparCGDevice extends DeviceWithState<State> implements IDevice {
 				// create a channel in state if necessary, or reuse existing channel
 				const channel = caspar.channels[mapping.channel] || { channelNo: mapping.channel, layers: {} }
 				channel.channelNo = Number(mapping.channel) || 1
-				channel.fps = 25
+				channel.fps = this.initOptions ? this.initOptions.fps || 25 : 25
```
gundelsby, line 483: Is this project too old to use the optional chaining? If not, this (and similar lines below) reads easier as `channel.fps = this.initOptions?.fps || 25`, no?

baltedewit: Yes, the version of Typescript used does not have optional chaining yet

src/devices/__tests__/casparcg.spec.ts
baltedewit: Yeap, I've changed these in `casparcg.spec.ts` now but when we add better linting we'll have to do this everywhere

PR #172, opened 2021-01-04
fix: don't set lockahead on a port that is playing

sparkpunkd:

A fix to quantel-specific behaviour of not resetting a clip when playing through a segment for the 2nd time after using setAsNext.

* Segment with a camera part followed by 2 (or more) Quantel VT parts. 
* Play camera, first VT, second VT ... set segment as next while playing 2nd VT. Take.
* Back on camera, then first VT plays OK. 2nd VT is either still playing or paused in preview. Take into 2nd VT and it is either still playing or paused on last frame.

Segment plays as expected.

* **Other information**:


sparkpunkd:
Related Playout Gateway PR: https://github.com/nrkno/tv-automation-playout-gateway/pull/72

sparkpunkd:
Pair programmed so assume reviewed by myself and @nytamin

PR #174, opened 2021-02-17
Fix: Quantel reload fragments if not found

nytamin:
This PR adds a retry-logic for the Quantel device, to solve the issue when the Clip doesn't exist in the Quantel-verse at the time of loading of fragments.

### DIFF
--- a/src/devices/__tests__/quantelGatewayMock.ts
+++ b/src/devices/__tests__/quantelGatewayMock.ts
@@ -14,7 +14,7 @@ export function setupQuantelGatewayMock () {
 		requestReturnsOK: true,
 		ignoreConnectivityCheck: false,
 		ISAOptionHasBeenProvided: false,
-		noClipsFond: false,
+		noClipsFound: false,
 		port: {
 			'my_port': {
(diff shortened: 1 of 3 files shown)


src/devices/__tests__/quantelGatewayMock.ts
```diff
@@ -14,6 +14,7 @@ export function setupQuantelGatewayMock () {
 		requestReturnsOK: true,
 		ignoreConnectivityCheck: false,
 		ISAOptionHasBeenProvided: false,
+		noClipsFond: false,
```
baltedewit, line 17: Typo?

baltedewit (approved):
This looks good to me but I can't make any judgements as to whether it will catch all edge cases so I will assume you have thought of them.

src/devices/__tests__/quantelGatewayMock.ts
nytamin: Typo indeed 🙂

sparkpunkd: Fixed.

PR #176, opened 2021-03-17
Feat/tv2 contribute

nytamin:
This PR contains various things from the olzzon fork that for some reason hasnt made it upstream yet.

A summary of the changes:
* Allow multiple layer mappings affecting one CasparCG layer. https://github.com/olzzon/tv-automation-state-timeline-resolver/commit/cb4bfdd6eede8f80e775c22bd43e12c09d1f491a
* Emitting new events `updateMediaObject` and `clearMediaObjects`, related to a feature that pipes back loaded-status of Viz-gfx (for stripes-display in the Sofie GUI). Note: This will likely be replaced by Package Manager later.
* VizMSE: fixes to preloading of elements.
* Sisyfos updates (@baltedewit could you review and elaborate on what those changes will mean?)


.gitignore
```diff
@@ -13,4 +13,5 @@ wallaby.conf.js
 .DS_Store
 src/types/coverage/
 .npmrc
+tv-automation-state-timeline-resolver
```
baltedewit, line 16: Not sure how this is relevant?

src/devices/casparCG.ts
```diff
@@ -511,7 +511,7 @@ export class CasparCGDevice extends DeviceWithState<State> implements IDevice {
 				const backgroundStateLayer = backgroundObj ? this.convertObjectToCasparState(mappings, backgroundObj, mapping, false) : undefined
 
 				if (foregroundStateLayer) {
-					channel.layers[mapping.layer] = {
+					channel.layers[layerName] = {
```
baltedewit, line 514: This change breaks the unit tests, and I don't think it's great idea in general because A) it can trip up casparcg-state and B) if I define layer1 to have { seek: 1 } and layer2 to have { seek: 2 } casparcg-state would probably send commands for both values at the same time which could lead to unexpected result.

baltedewit (changes requested):
I'll attempt a patch for the issue with `layerName` in the casparcg device. I can't say too much about the Viz integration but it looks OK to merge in general.

.gitignore
nytamin: I agree, it should probably not be in here.

It comes from https://github.com/olzzon/tv-automation-state-timeline-resolver/commit/016065b2dfb782c58f1976c428c213fc8ffabbce

@thomasslee97 Is this a remnant from a local file structure perhaps?

thomasslee97: This shouldn't be here for NRK

src/devices/casparCG.ts
ianshade: TV2 use case was addressed in a PR to this branch https://github.com/nrkno/tv-automation-state-timeline-resolver/pull/177

sparkpunkd (approved):
Good spot Balte on the CasparCG state issue and great that you added a test to cover the change 💯 

In general, I don't approve of just copying objects around to make things work, like with MediaObject arriving cut and paste into TSR. Code smell and over-reliance on 🦆 typing to solve and architectural flaw 😦. But ... as this is a temporary workaround until package manager resolves this issue in a architecturally sound way, I'm happy to approve this PR as a temporary workaround.

src/types/src/expectedPlayoutItems.ts
```diff
@@ -5,5 +5,7 @@ export type ExpectedPlayoutItemContent = ExpectedPlayoutItemContentVizMSE
 export interface ExpectedPlayoutItemContentBase {
 	/** Id of the rundown the items comes from */
 	rundownId: string
+	/** Id of the rundown playlist the items comes from */
+	playlistId: string
```
Julusian, line 9: This is not the appropriate place to define playlistId (or rundownId), as the blueprints will have to fill in something to get typescript to be happy. Playout gateway will then overwrite these values with real data.
This `ExpectedPlayoutItemContentBase` would be better to be kept in the non types side, or just within playout-gateway

PR #177, opened 2021-03-22
fix: extend templateData when allowing multiple mappings to reference 1 casparcg layer

ianshade:
On top of https://github.com/nrkno/tv-automation-state-timeline-resolver/pull/176/commits/49d313ba55086c367142efe1f180d6ea7d7c493a allows the `templateData` to be deep-extended if there are multiple mappings referencing one CasparCG layer.

`templateData` gets overwritten.

 If already exists, `templateData` is deep-extended.

* **Other information**:

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,6 +163,7 @@
     "casparcg-connection": "^5.1.0",
     "casparcg-state": "^2.1.0",
     "debug": "^4.3.1",
+    "deepmerge": "^4.2.2",
     "emberplus-connection": "^0.0.3",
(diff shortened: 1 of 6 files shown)


Julusian:
This isnt your fault, but I think it would be a good idea while doing this to replace the 'underscore-deep-extend' with something more recent and popular (and less underscore dependent is always good). Perhaps https://www.npmjs.com/package/deepmerge which we are now using in core?

ianshade:
Switched to `deepmerge`, except `atem.ts`, where I could not get the types right

PR #178, opened 2021-03-31
feat: Mapping layer name

thomasslee97:

Adds the `layerName` property to layer mappings, and utilises this in the Sisyfos integration as the default channel label.

No feature.

This PR adds an optional property to all mappings called `layerName`, which aims to serve two purposes.

First, this allows layer mappings to have a "friendly name" in core, something easier to recognise than the layer id used by blueprints.

Secondly, this PR adds a feature to the Sisyfos integration where, if the `layerName` is not blank, it will be used as the default label for a fader. This means that fader labels do not need to be defined by timeline objects in a studio / rundown baseline, however the default label can be overriden using timeline objects in the same way as the current behaviour. This also has the advantage that a system administrator can change the label of a sisyfos fader through the mapping settings, rather than via blueprints - reducing the need to touch code / reload NRCS data. This could also mean that a route set being activated could change the fader label without needing to rerun blueprints.

This approach could also be applied in the future to set the labels of ATEM sources, display some debug text on a caspar layer, or control the name of a camera in a tally system.

* **Other information**:

### DIFF
--- a/src/devices/__tests__/sisyfos.spec.ts
+++ b/src/devices/__tests__/sisyfos.spec.ts
@@ -40,19 +40,22 @@ describe('Sisyfos', () => {
 			device: DeviceType.SISYFOS,
 			mappingType: MappingSisyfosType.CHANNEL,
 			deviceId: 'mySisyfos',
-			channel: 0
+			channel: 0,
+			setLabelToLayerName: false
 		}
 		let myChannelMapping1: MappingSisyfos = {
 			device: DeviceType.SISYFOS,
 			mappingType: MappingSisyfosType.CHANNEL,
 			deviceId: 'mySisyfos',
-			channel: 1
+			channel: 1,
+			setLabelToLayerName: false
 		}
 		let myChannelMapping2: MappingSisyfos = {
 			device: DeviceType.SISYFOS,
 			mappingType: MappingSisyfosType.CHANNEL,
 			deviceId: 'mySisyfos',
-			channel: 1
+			channel: 1,
+			setLabelToLayerName: false
 		}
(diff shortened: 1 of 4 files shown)


nytamin (commented):
Two notes from me:
* please add documentation for the added property in the interface
* should it be renamed to `label`?

thomasslee97:
I've added documentation. I don't have any strong feelings on `layerName` vs `label`

baltedewit:
> This also has the advantage that a system administrator can change the label of a sisyfos fader through the mapping settings, rather than via blueprints - reducing the need to touch code / reload NRCS data.

I intend this to work the other way around, the system admin should change the labels on the mixer which should cascade into Sisyfos and then Sofie controls the channel purely by using the channel name/label. We have found that manually inputting the channel numbers into the right mapping is a very tedious and error prone process and not at all preferable when the mixer already outputs the mappings.

We need to have both sides of this working _and_ be able to assure that if the user unknowingly changes a label (because they think it's just a label in Sofie) that doesn't muck about with the control of the device.

thomasslee97:
How about a mapping-level setting for Sisyfos, something like "Set fader label to layer name"?

baltedewit:
That sounds like a good idea and I'd be happy to merge the PR after that

thomasslee97:
@baltedewit it's now behind a setting on a per-mapping basis

olzzon:
@baltedewit justbe aware that a fader can have more than one mixer-channel connected. That´s why i ended up using Sofie as main label source

@thomasslee97 my preference is always to use "id" and "label" as vars, as a var "name" sometime can be id and sometimes a label. .

PR #179, opened 2021-04-20
fix: don't run out transition when changing clip in preview

sparkpunkd:

Big fix.

When a lookahead chantges on an off-air port, such as when cueing a BTS, the outTransition runs from one to the other.

Out transitions do not run between clips when the state is "notOnAir true" to "notOnAir" true.

* **Other information**:

Cueing BTSs appears to be delayed by 2 seconds (actually 1400ms plus load clip time) without this change.


nytamin (approved):
I clarified the doc and am now happy!

PR #180, opened 2021-04-22
feat: mono repo

Julusian:

workflow/code structure change

this repository is a 'secret' mono-repo, with the second project living hidden inside the first

this is now more explicitly a mono-repo, using lerna (similar to what we have at https://github.com/nrkno/tv-automation-server-core/tree/release33/packages)

This restructure is required to allow us to use https://github.com/nrkno/tv-automation-sofie-code-standard-preset, as the structure really confused the linter (or it might have been typescript when in strict mode).

* **Other information**:

No work has been put into ci yet, I am hoping to switch it to github actions at the same time as fixing it up.

The code-preset it not being used yet, to minimise the chance of conflicts as this could be sat around a while

Major version has been bumped to allow for node 10 support to be dropped as dependencies are starting to drop that


nytamin (commented):
Yeah, I think this is a good starting point.

In my opinion, the biggest reason to mono-repoify this is to split out the different devices into separate packages, perhaps it would be good to do that as a part of this PR also?

Julusian:
@nytamin great, I shall do a few more bits here so I can merge it and do another pr for using the code-standard-preset afterwards (to minimise chance of painful merge conflicts)

Yeah, that will be very possible with this, but that will require some additional discussion to decide what and how. If we start by merging this, then we can gradually split more packages out over time

PR #181, opened 2021-04-27
feat: initial OBS implementation

jstarpl:

This is a new feature

No support

This PR adds support for controlling the OBS streaming production software using [Palakis/obs-websocket](https://github.com/Palakis/obs-websocket).

* **Other information**:


src/devices/obs.ts
```diff
@@ -0,0 +1,740 @@
...
+	clearFuture (clearAfterTime: number) {
+		// Clear any scheduled commands after this time
+		this._doOnTime.clearQueueAfter(clearAfterTime)
+	}
+
+	async terminate () {
+		this._setDisconnected = true
+		this._doOnTime.dispose()
+		this._obs.disconnect()
+		return Promise.resolve(true)
```
Julusian, line 266: As you're inside an async function, this could be simple `return true`. Up to you :)

src/devices/obs.ts
```diff
@@ -0,0 +1,740 @@
...
+			streaming: undefined,
+			scenes: {},
+			sources: {}
+		}
+	}
+
+	/** Called by the Conductor a bit before a .handleState is called */
+	prepareForHandleState (newStateTime: number) {
+		// clear any queued commands later than this time:
+		this._doOnTime.clearQueueNowAndAfter(newStateTime + 0.1)
```
Julusian, line 218: why `newStateTime + 0.1`? Won't that cause commands to be duplicated, as they are about to regenerated for `newStateTime`.

Not always duplicated, this could happen:
* First resolve of some future time 500, queue `currentScene = 1`
* prepareForHandleState(500), clears nothing
* Timeline is changed
* New resolve of time 500, queue `currentScene = 2`

So we will be performing a `currentScene = 1` folloewed 0.1ms later by `currentScene = 2`. This will quite possible cause a single frame flash of 1, depending on obs internals.

src/devices/obs.ts
```diff
@@ -0,0 +1,740 @@
...
+									obsTlCurrentScene.content.sceneName
+							}
+						}
+						break
+					case MappingOBSType.CurrentTransition:
+						if (
+							tlObject.content.type ===
+							TimelineContentTypeOBS.CURRENT_TRANSITION
+						) {
+							let obsTlCurrentTransition = (tlObject as any) as TimelineObjOBSCurrentTransition
```
Julusian, line 352: what if this was a lookahead object via `lookaheadForLayer`? Those objects are essentially background layer objects, which here will change the 'foreground' `currentTransition` property.

Same concern with the rest of the mappings handling below

src/devices/obs.ts
```diff
@@ -0,0 +1,740 @@
...
+		return commands
+	}
+
+	private _diffStates (
+		oldState: OBSState,
+		newState: OBSState
+	): Array<OBSCommandWithContext> {
+		let commands: Array<OBSCommandWithContext> = []
+
+		commands = commands.concat(
```
Julusian, line 675: Feel free to ignore this style preference

I think it is nicer to either use `commands.push(...this._resolveThing())` (so that commands can be const) or use the spread operator while building the array
```
const commands = [
...this._resolveThing(),
...this._resolveThing(),
]
```

src/types/src/obs.ts
```diff
@@ -0,0 +1,192 @@
...
+			is_local_file?: boolean
+			local_file?: string
+			looping?: boolean
+		}
+	}, {
+		sourceType: 'dshow_input' | 'browser_source' | 'window_capture' | 'image_source'
+	}>
+}
+
+export interface TimelineObjOBSMute extends TimelineObjOBSBase {
```
Julusian, line 184: I am wondering if this should be `TimelineObjOBSSourceAudio` instead of being just for muted, as there are other audio properties that I assume can be set over websocket, and it might be nice to be able to group inside of single timeline objects

Julusian (changes requested):
I have a few small concerns in the comments, and a couple of style suggestions.

It also looks like there has been some fighting over formatting in a few files, but nothing to worry about there

src/devices/obs.ts
jstarpl: I don't know 😅 I just copied this from the vMix module... So, what you're saying is that it would be better to do this as `.clearQueueNowAndAfter(newStateTime)`? I'm just a babe in the woods.

src/types/src/obs.ts
jstarpl: Yeah, good point. I just took commands and converted them into timeline objects, but I agree that it makes sense to add volume here as well.

jstarpl: Actually, after a bit of a think, I think I'm gonna keep it that way. The other properties that are available are volume faders and channel assignments. While I guess it may be useful to automate channel assignments, I suspect these changes will be separate from Mute/Unmute and Volume operations. I also think volume manipulation should also probably (eventually) live in it's own timeline object. My reasoning behind this is that with mute, you can almost use a semi-Sisyphos workflow, just putting audio sources on/off air, while having full manual control over their volume. Having volume automation alongside muting will invite using volume for audio automation and effectively invite the sort of human-machine conflicts Sofie allows to avoid. So I'll be keeping mute in a separate object, as that seems to be to be more practical.

Julusian (approved):

PR #182, opened 2021-05-18
fix: multiple VizMSE device fixes

ianshade:
Bug fixes and minor features in the VizMSE device

  - `makeReady` waits for initializations, halting playlist activation for a couple seconds
  - active<->rehearsal purges elements
  - duplication of some elements occurs

  fixes above, and:

  - prevents from purging baseline expectedPlayoutItems
  - improves logging
  - reports Viz Engines that become unavailable
  - attempts to reload elements after an Engine is restart

* **Other information**:
depends on https://github.com/olzzon/v-connection/pull/14 and https://github.com/olzzon/v-connection/pull/18

### DIFF
--- a/packages/timeline-state-resolver/src/devices/vizMSE.ts
+++ b/packages/timeline-state-resolver/src/devices/vizMSE.ts
@@ -338,6 +338,7 @@ export class VizMSEDevice extends DeviceWithState<VizMSEState> implements IDevic
 	 * @param okToDestroyStuff Whether it is OK to do things that affects playout visibly
 	 */
 	async makeReady (okToDestroyStuff?: boolean, activeRundownPlaylistId?: string): Promise<void> {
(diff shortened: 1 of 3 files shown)


packages/timeline-state-resolver/src/devices/vizMSE.ts
```diff
@@ -36,16 +36,15 @@ import {
...
-	VizEngine
+	VElement
 } from 'v-connection'
 
 import { DoOnTime, SendMode } from '../doOnTime'
 
 import * as crypto from 'crypto'
 import * as net from 'net'
 import { ExpectedPlayoutItem } from '../expectedPlayoutItems'
+import * as request from 'request'
```
Julusian, line 47: It would be nice to not use request if possible as it is EOL, but I know we are using it elsewhere in this repo so I'm not going to block this on that

Julusian (approved):
I feel a bit lost looking at some of this, but it all looks reasonable to me

PR #185, opened 2021-06-15
feat: resend failing http commands

baltedewit:
Resends failed HTTP commands as long as the object stays alive
Changes the deprecated request library to the got library

HTTP commands that fail are not retried, this can cause errors on air when DNS fails.


packages/timeline-state-resolver-types/src/httpSend.ts
```diff
@@ -17,6 +17,9 @@ export interface HTTPSendOptions {
 	makeReadyCommands?: HTTPSendCommandContent[]
 	/** Whether a makeReady should be treated as a reset of the device. It should be assumed clean, with the queue discarded, and state reapplied from empty */
 	makeReadyDoesReset?: boolean
+
+	/** */
```
gundelsby, line 21: This comment could be slightly more wordy :)

gundelsby (approved):
LGTM, but add that explanation for resendTime in the interface :)

PR #188, opened 2021-08-30
fix: more VizMSE device fixes

ianshade:
Bug fixes

- Restarting the playout gateway causes preloading elements from all the playlists instead of only the active one (when appropriate setting is ticked)
- Going from rehearsal<->active causes elements to be purged and preloaded again
- When playout gateway is started before MSE, the connection is never established
- After a restart of VizEngine, elements are not preloaded again

* **Other information**:


jesperstarkar:
Merging this to an experimenta branch for further testing

PR #200, opened 2022-01-03
fix: add missing atem-type to TimelineObjAtemAny

nytamin:
Add missing type `TimelineObjAtemAudioChannel` to `TimelineObjAtemAny`

### DIFF
--- a/packages/timeline-state-resolver-types/src/__tests__/library.spec.ts
+++ b/packages/timeline-state-resolver-types/src/__tests__/library.spec.ts
@@ -1,5 +1,4 @@
 // These imports are pointed to what external libraries will import
-// eslint-disable-next-line
 import { DeviceType as Types_DeviceType, TSRTimeline, TimelineObjEmpty } from '../../dist'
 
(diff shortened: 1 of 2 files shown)


packages/timeline-state-resolver-types/src/__tests__/library.spec.ts
```diff
@@ -1,4 +1,5 @@
 // These imports are pointed to what external libraries will import
+// eslint-disable-next-line
```
baltedewit, line 2: This hasn't come up in the CI as an issue so I don't think there should be an ignore here.

baltedewit (changes requested):

packages/timeline-state-resolver-types/src/__tests__/library.spec.ts
nytamin: I added this line due to getting a linting error (`no-extraneous-dependencies`) locally in the pre-commit script.
But you're right, it must have been something in my local setup, as I can't reproduce the issue on another system.

PR #201, opened 2022-01-04
Fix: Optimizations

nytamin:
This PR does a few minor fixes with the aim of increasing performance.

### DIFF
--- a/packages/timeline-state-resolver/src/conductor.ts
+++ b/packages/timeline-state-resolver/src/conductor.ts
@@ -72,6 +72,8 @@ export interface ConductorOptions {
 	multiThreadedResolver?: boolean
 	useCacheWhenResolving?: boolean
 	proActiveResolve?: boolean
+	/** When set, some optimizations are made, intended to only run in production */
+	optimizeForProduction?: boolean
 }
 interface TimelineCallback {
 	time: number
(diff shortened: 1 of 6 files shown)


packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -896,26 +909,21 @@ export class Conductor extends EventEmitter<ConductorEvents> {
 
 			// Push state to the right device:
 			await this._mapAllDevices(false, async (device: DeviceContainer<DeviceOptionsBase<any>>): Promise<void> => {
+				if (this._options.optimizeForProduction) {
```
baltedewit, line 912: Wouldn't it be better to use the conventional `NODE_ENV` environment variable for this?

nytamin: Ideally, yes it would. I intended this PR to be a minimal change, so consider this being a feature flag (because it botches the unit tests that are testing the abstract device).

If this turns out to be a good solution I believe that we should eventually remove the ABSTRACT devices altogether (ie keep the timeline-objects, but never spawn devices).

baltedewit: Feel free to merge then.

PR #209, opened 2022-03-23
The Big Sync (TV 2 -> release42)

ianshade:

* **Other information**:


baltedewit (approved):

packages/timeline-state-resolver-types/src/vizMSE.ts
```diff
@@ -160,6 +166,35 @@ export interface TimelineObjVIZMSEClearAllElements extends TSRTimelineObjBase {
 
 		/** Names of the channels to send the special clear commands to */
 		channelsToSendCommands?: string[]
+
+		/** IDs of the Show to use for taking the special template */
+		showId: string
```
nytamin, line 171: I'm slightly out of my depth here, but if I've understood things correctly, NRK aren't currently using the "shows"-concept.

Can this (and other properties and features that depend on it) be made optional? Or is there another solution for us?

ianshade: We have two types of elements: Internal, and External (Pilot). Internal elements require a Show, External elements don't. NRK is not using Internal elements, otherwise it would need to pass ShowIds as well.

ianshade: The ShowId used to be set in the device settings, hence we were to use only one Show. Now we started supporting playing out elements from more than one Show at the same time. If you were to start using Internal elements, you would need to make a blueprint setting for they showId(s).

nytamin: Ok, thanks for the clarification!

nytamin (approved):

PR #212, opened 2022-04-13
fix(vmix): account for start time when seeking an input

alvancamp:

Bug fix

Say that a video was supposed to have begun playing 5 seconds ago. Before this fix, TSR wouldn't account for that and would always start playing the media at timecode zero.

Now, TSR accounts for the start time of the instance when seeking a media input. If a media file was supposed to have started playing 5 seconds ago, TSR will now make VMix seek to 5 seconds into that media file when playback begins.

* **Other information**:

A new test has been added for this case.

### DIFF
--- a/packages/timeline-state-resolver/src/devices/__tests__/vmix.spec.ts
+++ b/packages/timeline-state-resolver/src/devices/__tests__/vmix.spec.ts
@@ -760,6 +760,133 @@ describe('vMix', () => {
 		expect(deviceErrorHandler).toHaveBeenCalledTimes(0)
 	})
 
+	test('Input seeking based on start time', async () => {
+		let device: any = undefined
+		const commandReceiver0 = jest.fn((...args) => {
+			// pipe through the command
+			return device._defaultCommandReceiver(...args)
+		})
+
+		const myLayerMapping0: MappingVMixInput = {
+			device: DeviceType.VMIX,
+			mappingType: MappingVMixType.Input,
+			deviceId: 'myvmix',
+			index: 2,
+		}
+		const myLayerMapping: Mappings = {
+			vmix_media0: myLayerMapping0,
+		}
+
+		const myConductor = new Conductor({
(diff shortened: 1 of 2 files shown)


nytamin (approved):

baltedewit (changes requested):
I see 2 issues with this

- The calculation does not take looping clips into account, i.e. a 10 second looping clip will be seeked past 10 seconds if it's on air for more than 10 seconds
- We should only send a position commands if the startTime changes

Good to have added unit testing for this!

alvancamp:
Thanks for the review!

I'm a little uncertain how I'd fix these issues. For the looping issue, wouldn't I need to know the actual duration of the file so that I could use it in a modulo operator with the computed position value? If so, how would I get that information?

For only sending the command if the startTime changes, do I need to add some additional internal state prop to the `VMixDevice` class to keep track of that?

baltedewit:
Sorry for replying late on this.

> For the looping issue, wouldn't I need to know the actual duration of the file so that I could use it in a modulo operator with the computed position value? If so, how would I get that information?

Yes, that's right. Said information could either come from vmix (if it exposes that?) or from the timeline object, adding a length property if it doesn't already exist. If the length property is missing just assume that the timing doesn't matter (like for background loops) and do not seek

> For only sending the command if the startTime changes, do I need to add some additional internal state prop to the VMixDevice class to keep track of that?

Yes, you would add a startTime property to the internal state to be able to compare these properties

alvancamp:
> Said information could either come from vmix (if it exposes that?) or from the timeline object, adding a length property if it doesn't already exist.

vMix does indeed expose the duration of media inputs, so that can be used in these calculations:

```ts
const duration = deviceState.reportedState.inputs[vmixMapping.index].duration
```

How would I test this though? Is there a way to tell TSR under test to use a hand-written mock reportedState for a device?

alvancamp:
Closing in favor of #235

PR #213, opened 2022-08-14
feat(Hyperdeck): add support for play and basic goto commands

alvancamp:

Feature

The `play` and `goto` commands of the Hyperdeck protocol are not supported.

`play` is fully supported. `goto` has basic support for selecting a clip by ID. Other aspects of `goto` have not been implemented.

* **Other information**:

Tests have been added and the functionality has been tested via quickTSR over a remote connection to a Hyperdeck unit. Here are the input files I used: https://gist.github.com/alvancamp/320d1dc7c4387bb031acf8e7e0a1edc8

### DIFF
--- a/packages/timeline-state-resolver-types/src/hyperdeck.ts
+++ b/packages/timeline-state-resolver-types/src/hyperdeck.ts
@@ -82,8 +82,16 @@ export interface TimelineObjHyperdeckTransport extends TimelineObjHyperdeck {
 		deviceType: DeviceType.HYPERDECK
 		type: TimelineContentTypeHyperdeck.TRANSPORT
 
-		/** The status of the hyperdeck. To start a recording, set to TransportStatus.RECORD */
+		/** The status of the hyperdeck. To start a recording, set to TransportStatus.RECORD. To start playback, set to TransportStatus.PLAY. */
 		status: TransportStatus
+		/** How fast to play the currently-playing clip [-5000 - 5000]. Default 1x speed is 100. 0 is stopped. Negative values are rewind. Values above 100 are fast-forward. */
+		speed: number
+		/** Whether or not to loop the currently-playing clip */
+		loop: boolean
+		/** Wheter or not to stop playback when the currently-playing clip is finished */
+		singleClip: boolean
+		/** The numeric ID of the clip to play. null = no clip */
+		clipId: number | null
 		/** The filename to record to */
 		recordFilename?: string
 	}
--- a/packages/timeline-state-resolver/src/devices/hyperdeck.ts
+++ b/packages/timeline-state-resolver/src/devices/hyperdeck.ts
@@ -41,6 +41,10 @@ export interface HyperdeckCommandWithContext {
 
 export interface TransportInfoCommandResponseExt {
 	status: TransportStatus
+	speed: number
+	singleClip: boolean
+	loop: boolean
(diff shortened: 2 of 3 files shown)


packages/timeline-state-resolver-types/src/hyperdeck.ts
```diff
@@ -82,8 +82,16 @@ export interface TimelineObjHyperdeckTransport extends TimelineObjHyperdeck {
 		deviceType: DeviceType.HYPERDECK
 		type: TimelineContentTypeHyperdeck.TRANSPORT
 
-		/** The status of the hyperdeck. To start a recording, set to TransportStatus.RECORD */
+		/** The status of the hyperdeck. To start a recording, set to TransportStatus.RECORD. To start playback, set to TransportStatus.PLAY. */
```
baltedewit, line 85: I think at this point we should create a typescript interface for each transport state, each with the correct properties and then `TimelineObjHyperdeckTransport` should be a union type of these interfaces. This way when somehow uses `TransportStatus.Play` they will not get type completions for recordFilename etc.

packages/timeline-state-resolver-types/src/hyperdeck.ts
```diff
@@ -82,8 +82,16 @@ export interface TimelineObjHyperdeckTransport extends TimelineObjHyperdeck {
...
 		status: TransportStatus
+		/** How fast to play the currently-playing clip [-5000 - 5000]. Default 1x speed is 100. 0 is stopped. Negative values are rewind. Values above 100 are fast-forward. */
+		speed: number
```
baltedewit, line 88: The `speed`, `loop` and `singleClip` should be optional with defaults of `100`, `false` and `true` since in most cases there will be no need to use anything else

packages/timeline-state-resolver/src/devices/hyperdeck.ts
```diff
@@ -313,11 +317,19 @@ export class HyperdeckDevice extends DeviceWithState<DeviceState, DeviceOptionsH
 								deviceState.transport = {
 									status: hyperdeckObjTransport.content.status,
 									recordFilename: hyperdeckObjTransport.content.recordFilename,
+									speed: hyperdeckObjTransport.content.speed,
```
baltedewit, line 320: Add aforementioned defaults here...

baltedewit (changes requested):
I'm quite happy with this PR, would like to see some of the types changed though.

alvancamp:
Thanks for the feedback! I _think_ I implemented the desired changes correctly. The union types for most of the transport states currently don't have any properties other than `status` because they aren't fully supported yet. Let me know if you'd like me to fully flesh these out even though they aren't supported in the actual device code yet.

alvancamp:
Might want to hold off on merging this. I'm testing it out with some live hardware via SuperConductor and seeing some bugs. I'm not yet sure if they are bugs in TSR or if they're bugs in SuperConductor. I'll keep you posted.

baltedewit:
I think it makes sense to have the preview and stopped states, they also shouldn't be much time implementation wise I think. The other once only make sense for manual control I think so I don't think we should worry about those.

alvancamp:
This is now ready for another review @baltedewit 🙂

baltedewit (approved):

PR #215, opened 2022-08-25
Suppress connection warnings for UDP OSC devices

alvancamp:
Bug fix

UDP OSC devices always have a 'disconnected' status which can cause warnings to be displayed in various consumers of TSR.

UDP OSC devices are now always assumed to be connected, to suppress these warnings.

* **Other information**:
N/A


alvancamp:
Nevermind, I completely misread some code that led me to this false fix.

PR #218, opened 2022-09-01
fix: stabilize timeline object callbacks by preferring originalStart in the callBackId

jstarpl:

Bug fix

The callBackId can sometimes be a unstable, resulting in multiple callbacks where there should only be one.

The callBackId uses originalStart falling back to start. Also, timeline state time is forwarded into `_diffStateForCallbacks`, instead of `getCurrentTime()`, since they can possibly be different.

* **Other information**:

### DIFF
--- a/packages/timeline-state-resolver/src/conductor.ts
+++ b/packages/timeline-state-resolver/src/conductor.ts
@@ -1123,11 +1123,9 @@ export class Conductor extends EventEmitter<ConductorEvents> {
 	}
 
 	private _diffStateForCallbacks(activeObjects: TimelineCallbacks, tlTime: number) {
-		const time = this.getCurrentTime()
-
 		// Clear callbacks scheduled after the current tlState
 		for (const [callbackId, o] of Object.entries(this._sentCallbacks)) {
-			if (o.time >= time) {
+			if (o.time >= tlTime) {
 				delete this._sentCallbacks[callbackId]
 			}
 		}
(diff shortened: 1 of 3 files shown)


nytamin (approved):

packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -1122,7 +1122,7 @@ export class Conductor extends EventEmitter<ConductorEvents> {
 		)
 	}
 
-	private _diffStateForCallbacks(activeObjects: TimelineCallbacks) {
+	private _diffStateForCallbacks(activeObjects: TimelineCallbacks, tlTime: number) {
 		const sentCallbacks: TimelineCallbacks = this._sentCallbacks
 		const time = this.getCurrentTime()
```
LindvedKrvang, line 1127: Do you still need this since you now parse the time in as a parameter?

jstarpl: The way I see it, this is _different_ time. `tlTime` is the time position at which the state has been evaluated. `time` is the current time. That's why callbacks should be cleared, based on a comparison with `time` in L1131, but `_queueCallback` should be at `tlTime` because that's when we want the callback to happen. @nytamin unless you see it differently? I might be misunderstanding the code here.

nytamin: I'm actually leaning towards using `tlTime` on L1131, and not using `getCurrentTime` in this function at all.

My reasoning is this:
`_diffStateForCallbacks()` is queued to be executed at a certain time ([using doOnTime.queue()](https://github.com/nrkno/sofie-timeline-state-resolver/blob/fix/brokenReportedTimings/packages/timeline-state-resolver/src/conductor.ts#L1026)).
Therefore, _if_ `_diffStateForCallbacks()` is executed in time, the two times would be identical. If however, it is not executed in time, we should modify `this._sentCallbacks` using `tlTime`, because that would render the same result, reducing the risk of it being impacted by race-conditions.

PR #219, opened 2022-09-16
Feature: timeline datastore

baltedewit:
This PR introduces a concept known as the _Timeline Datastore_. The datastore is a key value store that can be referenced from the content of a timeline object. When the timeline is resolved the references are replaced by the current value of the reference. If the timeline object started after the last update to the Timeline Datastore it may use a default value instead. 

Proposed API for timeline objects:
```ts
const mixEffect0: TSRTimelineObj = {
    id: 'me0',
    enable: {
        start: Date.now(),
    },
    layer: 'me0',
    content: {
        deviceType: DeviceType.ATEM,
        type: TimelineContentTypeAtem.MixEffect,

        $references: {
            input: { // the datastore key
                path: 'me.input', // property to override
                overwrite: false, // whether the default should override if the object is newer than the datastore value
            }
        },

        me: {
            input: 1,
            transition: AtemTransitionStyle.Cut
        }
    },
}
```
Proposed format of the datastore:
```ts
const ds = {
    'input': {
        value: 2, // the value to place into the timeline object
        modified: Date.now(), // when the datastore value was last updated
    }
}
```

This is a relatively small change to the TSR API but it allows systems using the TSR to control devices using very small database updates instead of updating the entire timeline. It's usage is entirely optional.

### DIFF
--- a/packages/timeline-state-resolver/src/conductor.ts
+++ b/packages/timeline-state-resolver/src/conductor.ts
@@ -18,6 +18,7 @@ import {
 	ResolvedTimelineObjectInstanceExtended,
 	TSRTimeline,
 	DeviceOptionsBase,
+	TimelineDatastoreReferences,
 } from 'timeline-state-resolver-types'
 import { AtemDevice, DeviceOptionsAtemInternal } from './devices/atem'
 import { EventEmitter } from 'eventemitter3'
@@ -44,6 +45,7 @@ import * as PAll from 'p-all'
 import PTimeout from 'p-timeout'
 import { ShotokuDevice, DeviceOptionsShotokuInternal } from './devices/shotoku'
 import { endTrace, fillStateFromDatastore, FinishedTrace, startTrace } from './lib'
+import { Datastore } from 'timeline-state-resolver-types/src'
 
 export { DeviceContainer }
 export { CommandWithContext }
@@ -113,12 +115,6 @@ export interface StatReport {
 	timelineSize: number
 	timelineSizeOld: number
 	estimatedResolveTime: number
-}
-export interface Datastore {
-	[key: string]: {
-		value: any
-		modified: number
-	}
 }
 
 export type ConductorEvents = {
@@ -1084,7 +1080,9 @@ export class Conductor extends EventEmitter<ConductorEvents> {
 
 		// find all references to the datastore that are in this state
 		const dependencies: string[] = Object.values(state.layers).flatMap(({ content }) =>
-			Object.keys(content.$references || {})
+			Object.values(content.$references || {}).map((r: TimelineDatastoreReferences[any]): string => {
+				return r.datastoreKey
+			})
 		)
 
 		// store all states between the current state and the new state
--- a/packages/timeline-state-resolver/src/conductor.ts
+++ b/packages/timeline-state-resolver/src/conductor.ts
@@ -18,7 +18,6 @@ import {
 	ResolvedTimelineObjectInstanceExtended,
 	TSRTimeline,
 	DeviceOptionsBase,
-	TSRTimelineObjBase,
 } from 'timeline-state-resolver-types'
 import { AtemDevice, DeviceOptionsAtemInternal } from './devices/atem'
 import { EventEmitter } from 'eventemitter3'
@@ -44,7 +43,7 @@ import PQueue from 'p-queue'
 import * as PAll from 'p-all'
 import PTimeout from 'p-timeout'
 import { ShotokuDevice, DeviceOptionsShotokuInternal } from './devices/shotoku'
-import { endTrace, FinishedTrace, startTrace } from './lib'
+import { endTrace, fillStateFromDatastore, FinishedTrace, startTrace } from './lib'
 
 export { DeviceContainer }
 export { CommandWithContext }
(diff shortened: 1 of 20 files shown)


packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -1062,6 +1075,128 @@ export class Conductor extends EventEmitter<ConductorEvents> {
...
+
+					for (const s of toBeFilled) {
+						const filledState: typeof s.state = JSON.parse(JSON.stringify(s.state))
+						Object.values(filledState.layers).forEach(({ content, instance }) => {
+							if ((content as TSRTimelineObjBase['content']).$references) {
+								Object.entries((content as TSRTimelineObjBase['content']).$references || {}).forEach(([key, ref]) => {
+									const datastoreVal = this._datastore[key]
+									const set = (obj: Record<string, any>, path: string, val: any) => {
+										const p = path.split('.')
+										p.slice(0, -1).reduce((a, b) => a[b], obj)[p.slice(-1)[0]] = val
```
jstarpl, line 1174: Should this be in a try-catch? It seems like something that can crash the entire conductor.

jstarpl (changes requested):

packages/timeline-state-resolver/src/conductor.ts
baltedewit: Yeah, could have definitely crashed it. I've updated the `set` method now so that it can't access undefined properties anymore. I think this is more useful anyway because it means we do not require a default for every value

packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -1062,6 +1074,111 @@ export class Conductor extends EventEmitter<ConductorEvents> {
...
+		])
+
+		const filledState: typeof state = JSON.parse(JSON.stringify(state))
+		Object.values(filledState.layers).forEach(({ content, instance }) => {
+			if ((content as TSRTimelineObjBase['content']).$references) {
+				Object.entries((content as TSRTimelineObjBase['content']).$references || {}).forEach(([key, ref]) => {
+					const datastoreVal = this._datastore[key]
+					const set = (obj: Record<string, any>, path: string, val: any) => {
+						const p = path.split('.')
+						p.slice(0, -1).reduce((a, b) => (a[b] ? a[b] : (a[b] = {})), obj)[p.slice(-1)[0]] = val
```
nytamin, line 1105: This algorithm should be deduplicated with the one at line 1156

nytamin (changes requested):
I think the logics in general look good!
I left a few notes on docs & refactoring

packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -114,6 +114,12 @@ export interface StatReport {
 	timelineSizeOld: number
 	estimatedResolveTime: number
 }
+export interface Datastore {
+	[key: string]: {
+		value: any
+		modified: number
```
nytamin, line 120: This could use some documentation, what it is (a timestamp) and how it is used (together with `overwrite`).

Possibly also rename it to be more similar to the `overwrite`-property, since the two are closely related. perhaps "overwriteTimestamp"?

packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -1062,6 +1075,128 @@ export class Conductor extends EventEmitter<ConductorEvents> {
...
+							if ((content as TSRTimelineObjBase['content']).$references) {
+								Object.entries((content as TSRTimelineObjBase['content']).$references || {}).forEach(([key, ref]) => {
+									const datastoreVal = this._datastore[key]
+									const set = (obj: Record<string, any>, path: string, val: any) => {
+										const p = path.split('.')
+										p.slice(0, -1).reduce((a, b) => a[b], obj)[p.slice(-1)[0]] = val
+									}
+
+									if (datastoreVal !== undefined) {
+										if (ref.overwrite) {
```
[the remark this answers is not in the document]
nytamin: Ah, got it!
It is a great feature! Perhaps the documentation of the `overwrite` property could be improved to include that explanation?

(possibly a personal opinion:) I'm unsure about the naming of the property, "overwrite" doesn't quite explain it.. hm.. Perhaps rename it to "useNewest", "useLatest" or something?

packages/timeline-state-resolver/src/lib.ts
```diff
@@ -217,3 +220,39 @@ export function endTrace(trace: Trace): FinishedTrace {
...
+ * Set a value on an object from a .-delimited path
+ * @param obj The base object
+ * @param path Path of the value to set
+ * @param val The value to set
+ */
+const set = (obj: Record<string, any>, path: string, val: any) => {
+	const p = path.split('.')
+	p.slice(0, -1).reduce((a, b) => (a[b] ? a[b] : (a[b] = {})), obj)[p.slice(-1)[0]] = val
+}
+export function fillStateFromDatastore(state: TimelineState, datastore: Datastore) {
```
nytamin, line 234: Personal opinion: This should probably moved to a separate file, since it has dependencies to `conductor.ts` it creates a circular dependency between `conductor.ts` and `lib.ts`.

nytamin (changes requested):

baltedewit:
The issues mentioned in the review were fixed during a collaborative VS code session.

PR #221, opened 2022-09-29
fix: threaded class error handler

jstarpl:

Bugfix

There is no `error` event handler registered on the device threads, meaning that if a device throws a thread-level error, nothing is there to catch it and log it.

Add an event handler to catch the errors. This should also improve the stability, because an uncaught error in the threadedClass could crash the entire playout gateway.

* **Other information**:


Julusian (approved):
This doesn't appear to add a handler for the AsyncResolver thread, but for the devices it looks good

PR #223, opened 2022-10-11
fix: re-adds a check for clearAllOnMakeReady before doing so

jesperstarkar:
* **What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
Fix, re-adding a feature missing since May 2nd.

* **What is the current behavior? (You can also link to an open issue here)**
If Clear-commands are specified, they will run on each rundown activation.

Clear-commands will only run (if specified) AND the option clearAllOnMakeRady is set to true.

* **Other information:**
Without this change, the option clearAllOnMakeReady is useless/not being used anywhere, and should be removed.

### DIFF
--- a/packages/timeline-state-resolver/src/__mocks__/net.ts
+++ b/packages/timeline-state-resolver/src/__mocks__/net.ts
@@ -33,12 +33,15 @@ export class Socket extends EventEmitter {
 	// this.emit('close')
 	// this.emit('end')
 
-	public connect(port: number, host: string) {
+	public connect(port: number, host: string, cb?: () => void) {
 		// this._port = port
 		// this._host = host
 
 		if (this.onConnect) this.onConnect(port, host)
 		setTimeoutOrg(() => {
+			if (cb) {
+				cb()
+			}
 			this.setConnected()
 		}, 3)
 	}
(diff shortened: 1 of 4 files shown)


jstarpl (changes requested):
Would it be possible to add a test to verify that this setting takes effect. There already is a _vizMSE: clear all elements_ test case, which I suppose could be just copied and modified to verify.

jesperstarkar:
Tests added

packages/timeline-state-resolver/src/__mocks__/net.ts
```diff
@@ -33,13 +33,12 @@ export class Socket extends EventEmitter {
 	// this.emit('close')
 	// this.emit('end')
 
-	public connect(port, host, cb) {
+	public connect(port: number, host: string) {
```
jstarpl, line 36: A connectListener is part of the [socket.connect API](https://nodejs.org/docs/latest-v14.x/api/net.html#net_socket_connect): I think it would be good to keep it. Maybe then the Node 16 tests wouldn't fail? IDK.

jstarpl (changes requested):

packages/timeline-state-resolver/src/__mocks__/net.ts
jesperstarkar: cb re-added as optional param. I see the docs you point at says host is optional too, will change that too.

jstarpl (approved):
Looks, good, tests pass now on all platforms. Good job! Go ahead for that merge button.

PR #225, opened 2022-10-18
Fix: casparCG template transitions bug

nytamin:
Bug fix.

CasparCG Templates doesn't actually support transitions, but we have code paths that support it (but apparently they are buggy).

When playing a template object, and (accidentally) has set a transition for it, the template isn't played properly.
(when stopped, a PLAY "empty" is sent, not a CG STOP)

CG PLAY and CG STOP are sent, as they should

* **Other information**:

TODO

- [ ] Verify that this is actually the right thing to do, and that templates actually does not and should not support transitions


baltedewit:
There is an edge case here when your template doesn't have an out transition and you want to do a fade out using the casparcg transitions

PR #230, opened 2022-11-10
Fix: http get requests

nytamin:

Bug fix

HTTPSend device didn't send GET requests at all

HTTPSend now sends GET requests.

User can also choose how the `body` is sent ("json" or "form" type) for POST/PUT/etc requests.
  ref: https://github.com/sindresorhus/got/blob/main/documentation/2-options.md#json

* **Other information**:

I've tested this to work, using quickTSR and a local http-server.
This should be backwards-compatible as it doesn't change the default way data is sent for POST/PUT/etc.


baltedewit (approved):
I believe it is convention to use "params" for url parameters and "body" or "payload" for POST requests. Approved of the code changes but I think following this convention would be better.

PR #231, opened 2022-11-16
Add support for Node18

nytamin:
This PR adds support for Node v18 in CI and in "engines" in package.js.

Since it was such a small change, I thought that we could just add it right away.

FYI: I did the same in threadedclass, and[ it seems to work without a hitch](https://github.com/nytamin/threadedClass/actions/runs/3479117410), so that should be good too for v18.

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -71,7 +71,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x]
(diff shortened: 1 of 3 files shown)


Julusian (approved):
It would be a good idea to do a quick manual test with quickTSR to verify that it does work correctly. Nothing major, just something simple in caspar plays out when all threading is enabled

baltedewit:
Should be release48 I think?

Julusian:
@baltedewit maybe, but it is a fix and changes no code, so its not unreasonable to put it into 47. Theres no additional risk to the release by including this

nytamin:
I've tested it locally using quickTSR and Node v18, seems to work well!

PR #232, opened 2022-11-22
Feat: JSON schema manifests for actions

baltedewit:
This PR adds JSON schemas for action manifests. There is a script to compile interfaces from the JSON files into the types package.  There's also a new API method called `executeAction`. A script to extract strings for translations is included and will extract anything calling `t()` as well as relevant strings from the JSON files.

Eventually I hope to use similar JSON files to describe the mappings and device options so more parts of the TSR can use generated UI's.


nytamin (approved):
Looks good!

Before merging, it might be good to have the nn and nb-translations filled in, so we have something to play with later.

PR #233, opened 2022-12-01
Ci/update gha workflows

PeterC89:
Updating GHA workflows to remove depreciated actions

Workflows run but with depreciation warnings

Workflows continue to run but without depreciation warnings

* **Other information**:
The **yuya-takeyama/docker-tag-from-github-ref-action** has been replaced with some bash scripting that **should** do the same thing

### DIFF
--- a/.github/workflows/node.yaml
+++ b/.github/workflows/node.yaml
@@ -21,9 +21,9 @@ jobs:
         lib-name: [timeline-state-resolver, timeline-state-resolver-types]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Use Node.js
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version-file: '.node-version'
       - name: Prepare Environment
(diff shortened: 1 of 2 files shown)


PeterC89:
> I am looking at this, and wondering if using this action is a relic of how we used to do things..
> 
> We use this to help generate a npm prerelease tag such as `1.0.0-nightly-feat-i-changed-something-20221201-120102-abcdef12.0`, where the `feat-i-changed-something` portion comes from this action (but is also put through some extra processing on line 105). We don't care about the exact format, its to give the name some meaning to identify what it is and give it some extra uniqueness.
> 
> So what did this action add? It might just be too late in the week, but would we get a close enough result by simply using `${{ github.ref_name }}`? Instead of `latest` it will be `master`, but that is probably a good thing. I'm pretty sure `ref_name` didn't exist a while back, and we had to parse it out of `ref`, which could explain why we used this action.
> 
> I'm not opposed to the bash scripting, if it does add value

Good question, maybe something we should gather thoughts on?
Absolutely in favour of taking out the workflow if it's no longer required for sure.

PeterC89:
@Julusian How's it look now?
Taken all the bash scripting out

baltedewit (approved):
I think this PR achieves what we need right now

Julusian (approved):

PR #236, opened 2023-01-06
Feat: Service class & State Handler

baltedewit:
This PR contains a new structure for building device integrations. The main purpose is to move any duplicated logic from the device integrations into shared classes and by doing so make the Device classes only responsible for implementing the state diffs and communicating with the device.

The handling of states itself has also been rewritten to only calculate commands for the next state change instead of every planned state change at once.

The new API is compatible with the old DeviceContainer so it's transparent to Conductor class and we can gradually move more device integrations to this new structure.


packages/timeline-state-resolver/src/service/stateHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+			mappings: Mappings
+		) => DeviceState,
+		private diffDeviceStates: (oldState: DeviceState, newState: DeviceState, mappings: Mappings) => Command[],
+		private executeCommand: (command: Command) => any
+	) {
+		this.clock = setInterval(() => {
+			// main clock to check if next state needs to be sent out
+			const nextState = this.stateQueue[0]
+
+			if (this.currentState && nextState && nextState.state.time <= Date.now()) this.executeNextStateChange()
```
ianshade, line 33: This looks like a floating promise to me. And because this is running on an interval, is there no risk of executing the next state change twice if for any weird reason [this part](https://github.com/nrkno/sofie-timeline-state-resolver/pull/236/files#diff-a8cbc2cdbe840e77a7479a09a30f6f41e81435fdc8e36b3562bb6a70b44f0ff5R97-R99) does not resolve within 20ms and the interval fires again?

packages/timeline-state-resolver/src/service/stateHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+		for (let command of newState.commands) {
+			this.executeCommand(command) // should these be awaited? what if they time out?
+		}
+
+		this.currentState = newState as {
+			commands: Command[]
+			deviceState: DeviceState
+			state: Timeline.TimelineState<TSRTimelineContent>
+			mappings: Mappings
+		}
```
ianshade, line 119: This type is repeated, could it be extracted?

packages/timeline-state-resolver/src/service/stateHandler.ts
baltedewit: No, it shouldn't execute twice because it'll be removed from the state queue after the first execution. However, some other weirdness may ensue so I'll need to put in some logic to ensure we don't execute multiple `executeNextStateChange` at the same time

packages/timeline-state-resolver/src/service/stateHandler.ts
```diff
@@ -0,0 +1,220 @@
...
+		this.logger = context.logger
+
+		this.convertTimelineStateToDeviceState = (s, m) => device.convertTimelineStateToDeviceState(s, m)
+		this.diffDeviceStates = (o, n, m) => device.diffStates(o, n, m)
+		this.executeCommand = async (c) => device.sendCommand(c)
+
+		this.clock = setInterval(() => {
+			// main clock to check if next state needs to be sent out
+			const nextStates = this.stateQueue.filter((s) => s.state.time < Date.now() + 20) // any states to be executed before the next tick
+
```
Julusian, line 46: should this be using `Date.now()` or the `getCurrentTime()` method being passed around?

```suggestion
this.stateQueue.forEach((nextState) => {
	const nextTime = Math.max(0, nextState.state.time - Date.now())
	if (nextTime > 20) return // Ignore any states for a future tick

```
This avoids needing to do a similar calculate twice, and reduces the amount of array iterations. It also clarifies that `nextTime` can only ever be 0-20

packages/timeline-state-resolver/src/service/stateHandler.ts
```diff
@@ -0,0 +1,220 @@
...
+				setTimeout(() => {
+					if (!this._executingStateChange && this.stateQueue[0] === nextState) {
+						// if this is the next state, execute it
+						this.executeNextStateChange().catch((e) => {
+							this.logger.error('Error while executing next state change', e)
+						})
+					}
+				}, nextTime)
+			})
+		}, 20)
```
Julusian, line 58: This should be a named constant/class property, as this `20` is used in 2 places that will need to be updated together

packages/timeline-state-resolver/src/service/stateHandler.ts
```diff
@@ -0,0 +1,220 @@
...
+		clearInterval(this.clock)
+		this.stateQueue = []
+	}
+
+	async clearFutureStates() {
+		this.stateQueue = []
+	}
+
+	async handleState(state: Timeline.TimelineState<TSRTimelineContent>, mappings: Mappings) {
+		const nextState = this.stateQueue[0]
```
Julusian, line 71: what if `this.stateQueue` is empty?

packages/timeline-state-resolver/src/service/stateHandler.ts
```diff
@@ -0,0 +1,220 @@
...
+		this.currentState = {
+			commands: [],
+			deviceState: state,
+			state: this.currentState?.state || { time: Date.now(), layers: {}, nextEvents: [] },
+			mappings: this.currentState?.mappings || {},
+		}
+		await this.calculateNextStateChange()
+	}
+
+	clearFuture(t: number) {
```
Julusian, line 107: I think this naming is unclear. There is already a `clearFutureStates`, and this `clearFuture` is very different.
Perhaps `clearStatesBeforeTime`?

packages/timeline-state-resolver/src/service/stateHandler.ts
```diff
@@ -0,0 +1,220 @@
...
+				.then(() => {
+					if (newState.measurement) this.context.reportStateChangeMeasurement(newState.measurement.report())
+				})
+				.catch((e) => {
+					this.logger.error('Error while executing next state change', e)
+				})
+		}
+
+		this.currentState = newState as ExecutedStateChange<DeviceState, Command>
+		this._executingStateChange = false
```
Julusian, line 196: the execution of commands is not being awaited, so this will be set while the commands are still executing. Is this intentional?

packages/timeline-state-resolver/src/service/stateHandler.ts
```diff
@@ -0,0 +1,220 @@
...
+			// the next state changed
+			if (nextState) nextState.commands = undefined
+			this.calculateNextStateChange().catch((e) => {
+				this.logger.error('Error while executing next state change', e)
+			})
+		}
+	}
+
+	async setCurrentState(state: DeviceState) {
+		this.currentState = {
```
Julusian, line 98: what happens if this is called while commands are being executed? Could execution of `executeNextStateChange()` overwrite this before it gets processed?

packages/timeline-state-resolver/src/integrations/osc/index.ts
```diff
@@ -340,14 +184,33 @@ export class OSCMessageDevice extends DeviceWithState<OSCDeviceState, DeviceOpti
...
+
+	private _oscSender(msg: osc.OscMessage, address?: string | undefined, port?: number | undefined): void {
+		this.emit('debug', 'sending ' + msg.address)
 		this._oscClient.send(msg, address, port)
 	}
 	private runAnimation() {
 		for (const addr in this.transitions) {
 			// delete old tweens
-			if (this.transitions[addr].started + this.transitions[addr].transition!.duration < this.getCurrentTime()) {
+			if (this.transitions[addr].started + this.transitions[addr].transition!.duration < Date.now()) {
```
Julusian, line 213: Do we not want `this.getCurrentTime()` anymore?

Julusian (commented):
I have some concerns about the async flow in some areas particularly stateHandler.

I do like how the osc implementation looks now, I don't see anything there which looks like it is duplicated across all the integrations

Ping me if you want to talk through the async flow, perhaps we could document it together to help me understand

packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -615,9 +607,24 @@ export class Conductor extends EventEmitter<ConductorEvents> {
...
+				try {
+					newDevice = await RemoteDeviceInstance.create(
+						'../../dist/service/DeviceInstance.js',
+						'DeviceInstanceWrapper',
+						deviceId,
+						deviceOptions,
+						() => 0,
+						{
+							...threadedClassOptions,
+							threadUsage: deviceOptions.isMultiThreaded ? 0.1 : 0,
```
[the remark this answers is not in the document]
baltedewit: I think I took a copy paste from the deviceOptions of the abstract device, I don't know how this could ever be correct

packages/timeline-state-resolver/src/service/devices.ts
```diff
@@ -0,0 +1,21 @@
+import { OscDevice } from '../integrations/osc'
+import { DeviceType } from 'timeline-state-resolver-types'
+import { Device } from './device'
+
+export interface DeviceEntry {
```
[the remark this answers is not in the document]
baltedewit: I don't know what it does, it looks like it never has had documentation. I think it signifies that a device integration doesn't keep a connection open (like the http devices). I've tried to sort of reduce the amount of breaking changes as much as I could and needed to put it somewhere but if we really can't figure out where it's used maybe we should just remove it.

packages/timeline-state-resolver/src/service/stateHandler.ts
baltedewit: Yes, this exists to wait for the next state to have been removed from the state queue and set as the current state. Although in my initial implementation I assumed all commands to be sent as salvos. Perhaps now that a sequential mode exists it would be better to await the last command to be sent in that mode.

packages/timeline-state-resolver/src/integrations/osc/index.ts
baltedewit: I'm undecided. I'm not convinced we used this.getCurrentTime in any meaningful way before but I know a lot of the Date.now usage just snuck in while typing away.

packages/timeline-state-resolver/src/service/remoteService.ts
```diff
@@ -0,0 +1,151 @@
...
+export type DeviceContainerEvents = {
+	error: [context: string, err: Error]
+}
+
+/**
+ * A device container is a wrapper around a device in ThreadedClass class, it
+ * keeps a local property of some basic information about the device (like
+ * names and id's) to prevent a costly round trip over IPC.
+ */
+export class RemoteService<TOptions extends DeviceOptionsBase<any>> extends EventEmitter<DeviceContainerEvents> {
```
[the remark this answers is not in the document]
jstarpl: I think it's just an established NodeJS design pattern: https://nodejs.org/api/events.html#events. With `eventemitter3` it makes type-safe event emitters much quicker to write and maintain.

packages/timeline-state-resolver/src/service/stateHandler.ts
baltedewit: Do you want to integrate a promise queue in here like we do in the main conductor? The only worry about race conditions that I have is when the `calculateNextStateChange` is called from the `executeNextStateChange` _and_ it takes a long time to execute for some reason. I don't even think this should ever happen so perhaps we should back out of the executing when it does and wait for another scheduling tick.

packages/timeline-state-resolver/src/service/stateHandler.ts
baltedewit: nextState will be undefined and the new state will become the next state

packages/timeline-state-resolver/src/service/remoteDeviceInstance.ts
```diff
@@ -0,0 +1,168 @@
...
+					// This is called if a child crashes
+					if (container.onChildClose) container.onChildClose()
+				}),
+				ThreadedClassManager.onEvent(container._device, 'error', (error) => {
+					container.emit('error', `${orgClassExport} "${deviceId}" threadedClass error`, error)
+				}),
+			]
+		}
+
+		await container.reloadProps()
```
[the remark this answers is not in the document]
baltedewit: although not directly related to this PR, I agree and will change this in the original DeviceContainer class as well

packages/timeline-state-resolver/src/service/remoteDeviceInstance.ts
```diff
@@ -0,0 +1,168 @@
...
+
+export type DeviceContainerEvents = {
+	error: [context: string, err: Error]
+}
+
+export abstract class BaseRemoteDeviceIntegration<
+	TOptions extends DeviceOptionsBase<any>
+> extends EventEmitter<DeviceContainerEvents> {
+	protected abstract _device: ThreadedClass<DeviceInstanceWrapper> | ThreadedClass<Device<TOptions>>
+	protected _deviceId = 'N/A'
```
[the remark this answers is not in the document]
baltedewit: This should all be moved back into one class when we get rid of the old Device classes, let's revisit it by then.

packages/timeline-state-resolver/src/service/remoteService.ts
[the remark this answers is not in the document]
Julusian: I remember that in c# you could define in your class `public event CommandHandler OnReceive;`, and then do `myClass.OnReceive += myHandler` to add a handler to that emitter.
I have not seen an analogue to that in the javascript world. Maybe this is something rxjs could do, but I have no experience with that library.

As Jan says, the normal way to do events in nodejs is to make the class extend EventEmitter. But the builtin version lacks type safety of the events, so we are using `eventemitter3` instead to provide us with the type safety.

packages/timeline-state-resolver/src/service/remoteDeviceInstance.ts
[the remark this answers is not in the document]
Julusian: Sounds reasonable. Would be nice to have a line somewhere documenting that this is what you intend to do

PR #237, opened 2023-01-19
feat!: json schemas for device config and mappings

Julusian:

breaking change

The options and mappings types for devices are handwritten typescript interfaces.

These types are now generated from json-schemas.
The schemas are then used in Sofie to generate the UI for user input.

Some of the breaking changes are not strictly necessary, but make the generated types be more consistent and easier to work with.

1) The structure of `Mapping` has changed. It has an `options: TOptions` property, which is a generic for the specific mapping type. It could be possible to avoid this, but that would result in more complex and confusing types and this new structure matches what we are doing elsewhere in Sofie.
1) Mappings are now required to have a `mappingType` property. In many cases, there is a signle mapping type, so this has a single value. It could be made so that this is not the case, but that adds more complexity in other places.
1) Various mapping types were using numeric enum values. These are now required to be string enums. Trying to avoid this could complicate the type generation, and would likely result in string vs number issues, as the json schema would want them to be a string

* **Other information**:

Enums which have changed to strings:
- MappingAtemType
- MappingOBSType
- MappingPanasonicPtzType
- MappingVMixType

Newly added enums for existing mapping types:
- MappingCasparCGType
- MappingHyperdeckType
- MappingQuantelType
- MappingSingularLiveType
- MappingSofieChefType


Julusian:
> The structure of Mapping has changed. It has an options: TOptions property

I am wondering if I should undo portion of the change, as its going to be the trickier one to handle in Sofie (some complex migrations).
I think it would be easier than I originally thought to handle here. Each mapping could be generated as `MappingThingInternal`, with an autogenerated wrapper of `export type MappingThing = Mapping & MappingThingInternal`.

packages/timeline-state-resolver/scripts/schema-types.mjs
```diff
@@ -8,22 +8,59 @@ import * as path from 'path'
...
+	})
+
+	await fs.writeFile('../timeline-state-resolver-types/src/generated/action-schema.ts', BANNER + '\n' + actionSchema)
+} catch (e) {
+	console.error('Error while generating action-schema.json, continuing...')
+	console.error(e)
+}
 
-const PrettierConf = JSON.parse(await fs.readFile('../../node_modules/@sofie-automation/code-standard-preset/.prettierrc.json'))
+// convert action-schema
```
baltedewit, line 35: This should be common-options

packages/timeline-state-resolver/scripts/schema-types.mjs
```diff
@@ -8,22 +8,59 @@ import * as path from 'path'
...
+try {
+	const commonOptionsDescr = JSON.parse(await fs.readFile('./src/$schemas/common-options.json'))
+	const actionSchema = await compile(commonOptionsDescr, 'DeviceCommonOptions', {
```
baltedewit, line 38: Same here

baltedewit (approved):

PR #238, opened 2023-02-03
SOFIE-1282: VMix TCP connection

baltedewit:
This PR replaces the vmix http connection for a TCP connection since it is known to be a bit more reliable.


packages/timeline-state-resolver/src/devices/vmixAPI.ts
```diff
@@ -1,91 +1,200 @@
...
 			}
-			this._options = options
 		}
 
-		return new Promise((resolve) => {
-			this.once('initialized', () => {
-				this.emit('stateChanged', this.state)
-				resolve(true)
-			})
+		this._socket = new Socket()
```
jstarpl, line 156: I think this should use `this._socket.setNoDelay(true)` (https://nodejs.org/api/net.html#socketsetnodelaynodelay)

packages/timeline-state-resolver/src/devices/vmixAPI.ts
```diff
@@ -1,91 +1,206 @@
...
+					}
+					response.body = lines.join('\r\n')
+					processedLines += lines.length
+				}
+
+				// now do something with response
+				this.emit('data', response)
 			} else {
-				this.emit('disconnected')
+				// well this is not happy, do we do something?
```
nytamin, line 134: If this happens, should we at least emit a warning?

packages/timeline-state-resolver/src/devices/vmixAPI.ts
```diff
@@ -1,91 +1,206 @@
...
+				processedLines++
+
+				// parse additional lines if needed
+				if (!Number.isNaN(responseLen)) {
+					let len = responseLen
+					const lines: string[] = []
+
+					while (len > 0) {
+						const l = this._unprocessedLines[lines.length + 1] // offset of 1 because first line is not data
+						if (!l) break // todo - should we wait for streaming data to come in?
```
nytamin, line 122: todo comment

packages/timeline-state-resolver/src/devices/__tests__/vmixAPI.spec.ts
```diff
@@ -10,48 +9,50 @@ async function wait(time = 1) {
 	})
 }
 describe('vMixAPI', () => {
```
nytamin, line 11: Perhaps we should add a test in here that tests the functinoality in `BaseConnection._processIncomingData()`, 
especially things related to messages-split-over-multiple-lines?

nytamin (approved):
Looks good after a read-through!

packages/timeline-state-resolver/src/devices/__tests__/vmixAPI.spec.ts
jstarpl: I've tweaked the behavior of the vMix mock so that it hits the messages split over multiple lines and messages split over multiple packets too, so I hope this is now handled.

PR #239, opened 2023-02-07
feat: NORA Core Device (SOFIE-1377)

jstarpl:

Feature

This implements a bespoke NORA Core device.

* **Other information**:


packages/timeline-state-resolver-types/src/noraNRK.ts
```diff
@@ -0,0 +1,62 @@
...
+
+	/** Minimum time in ms before a command is resent, set to <= 0 or undefined to disable */
+	resendTime?: number
+}
+
+export type TimelineObjNoraNRKAny = TimelineObjNoraNRKRequest
+export interface TimelineObjNoraNRKBase extends TSRTimelineObjBase {
+	content: {
+		deviceType: DeviceType.NORA_NRK
+		// type: TimelineContentTypeCasparCg
```
Julusian, line 54: It would be good to have a type field here. This is something that becomes required with the json schemas in r49

packages/timeline-state-resolver-types/src/noraNRK.ts
```diff
@@ -0,0 +1,62 @@
...
+import { TSRTimelineObjBase, DeviceType, TimelineDatastoreReferencesContent } from '.'
+
+export interface MappingNoraNRK extends Mapping {
+	device: DeviceType.NORA_NRK
+}
+
+export interface NoraNRKCommandContent {
+	group: string
+	groupSuffix?: string
+	channel: string
```
nytamin, line 11: Possibly move these to the Mapping instead?

nytamin (approved):
Notes after a talk with Jan:

The idea of this initial implementation is to have feature-parity with the old httpSend device. As such, I think this is a good-enough implementation :)

We also talked about the next step where the device would be listening to the actual state of Nora, and I also think that is a good implementation plan.

jstarpl:
Since the use case for this PR has been resolved in another way, the decision has been made to mothball this PR as a draft.

PR #240, opened 2023-02-14
feat(vmix): add support for ListRemoveAll and ListAdd commands

alvancamp:

This PR introduces support for the `ListRemoveAll` and `ListAdd` commands in vMix.

These commands are not supported.

If the new and optional `listFilePaths` string array is provided to a `TimelineContentVMixInput` timeline object, a few operations will occur:

1. `ListRemoveAll` will be dispatched to the mapped input to clear the list.
2. `ListAdd` will be dispatched once for each file in the `listFilePaths` array.

* **Other information**:
The file paths need to use Windows-style path separators and are automatically URI-encoded. Also, files added to a `List` input via the API tend to not show up in the vMix UI when inspecting the list. That is to say: the videos might be there and playing just fine, but the actual list will look empty in vMix.


packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -793,6 +794,27 @@ export class VMixDevice extends DeviceWithState<VMixStateExtended, DeviceOptions
 					timelineId: '',
 				})
 			}
+			if (input.listFilePaths !== undefined && !_.isEqual(oldInput.listFilePaths, input.listFilePaths)) {
+				commands.push({
```
Julusian, line 798: will this result in other properties of the list being reset too?
what happens if a clip from the list is playing when this is run?

packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -793,6 +794,27 @@ export class VMixDevice extends DeviceWithState<VMixStateExtended, DeviceOptions
 					timelineId: '',
 				})
 			}
+			if (input.listFilePaths !== undefined && !_.isEqual(oldInput.listFilePaths, input.listFilePaths)) {
```
Julusian, line 797: `input.listFilePaths !== undefined`
Is this correct?

When deactivating a rundown, the studio will go into the studio baseline state which will most likely not have a timeline object defining this list. So the clips wont be cleared? (Im assuming, I havent read enough to say for certain)

packages/timeline-state-resolver/src/integrations/vmix/index.ts
alvancamp: > will this result in other properties of the list being reset too?

No, other properties such as "sharpen" and "mirror" are retained.

> what happens if a clip from the list is playing when this is run?

It too is removed.

packages/timeline-state-resolver/src/integrations/vmix/index.ts
alvancamp: Good catch, that was indeed incorrect. I've edited it so that clips will be cleared when returning to a baseline which does not have a timeline object defining this list.

packages/timeline-state-resolver/src/integrations/vmix/index.ts
Julusian: > > what happens if a clip from the list is playing when this is run?
> 
> It too is removed.

Won't that be problematic?  
When playing a clip, and a new one is added to the list by a lookahead (or whatever causes it to be added to the list). Which results in the playing clip to be removed and re-added?

I assume it is the case that there is a list for each AB player. So maybe this scenario couldn't be hit, but I would want to have the next clip preloaded while the current one is playing to allow for seamless playback, or user confidence that the clip does load.

jstarpl: I think this is fine for now. I suppose there could be an option of individually removing the _removed_ items from that array, and then adding the ones added, but personally, I feel this could just be fixed by whoever needs it fixed.

jstarpl:
> I've just realized that vMix has a quirk that we could perhaps would around in TSR:
> 
> (...)
> 
> What do y'all think?

I think that if this ensures that vMix behaves predictably, I'm all for it. The TSR is in full control of this input, so if explicitly pausing it means that it will behave the same every time then that's good.

jstarpl (approved):

packages/timeline-state-resolver/src/integrations/vmix/index.ts
jstarpl: @Julusian are you satisfied with leaving this for now, or do you see a scenario where lookaheads _can_ wreak havoc here?

PR #243, opened 2023-02-21
Pull changes from TV2 Danmark, develop branch as of 2023-02-21

jstarpl:

* **Other information**:


packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -442,225 +500,216 @@ export class Conductor extends EventEmitter<ConductorEvents> {
...
 	 * @returns A promise that resolves with the initialised device, or rejects with an error message.
 	 */
 	public async initDevice(
 		deviceId: string,
 		deviceOptions: DeviceOptionsAnyInternal,
-		activeRundownPlaylistId?: string
+		activeRundownPlaylistId?: string,
+		options?: { signal?: AbortSignal }
 	): Promise<DeviceContainer<DeviceOptionsBase<any>>> {
+		const throwIfAborted = this.throwIfAborted.bind(this, deviceId, 'initialisation')
```
Julusian, line 709:
```suggestion
		const throwIfAborted = this.throwIfAborted.bind(this, options?.signal, deviceId, 'creation')
```

Missing a parameter. Functions like `.bind` and `.call` are not being type safe, but should be enabled in our config https://www.typescriptlang.org/tsconfig#strictBindCallApply

packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -1522,3 +1589,43 @@ function removeParentFromState(
...
+): Promise<T> {
+	const mainPromise = func(abortSignal)
+	if (!abortSignal) {
+		return mainPromise
+	}
+	let resolveAbortPromise: Function
+	const abortPromise = new Promise<void>((resolve, reject) => {
+		resolveAbortPromise = () => {
+			resolve()
+			// @ts-expect-error
```
Julusian, line 1611: These need a reason for why we expect an error. I can't tell why this would produce something that isnt easily fixable.

Julusian (approved):
I've not read much of the viz or tricast implementations, the rest looks good apart from a couple of small comments

packages/timeline-state-resolver/src/conductor.ts
Julusian: Oh, by 'strict' they mean this signature `bind<T, AX, R>(this: (this: T, ...args: AX[]) => R, thisArg: T, ...args: AX[]): (...args: AX[]) => R;` Which this call does fall into as the arguments do match `(string | AbortController | undefined)[]`.

So yeah, I don't like the use of `.bind()` here. I would rather see this as `const throwIfAborted = ( () => this.throwIfAborted(options?.signal, deviceId, 'initialisation')` so that we have real type safety

nytamin (approved):

Julusian:
I can't explain why this has only started happening with this merge, but this is happening because draining async queues when mocking time in jest is hard. Essentially the problem was that when sending in `BURST` mode, one command was being sent per `setTimeout`, with the next being queued in its own `setTimeout`. So to send all these commands means telling jest to execute the pending timers.  
The simpler option is to make `DoOnTime` be more efficient in how it sends, and execute all the sends in one go. I don't know if this will have any impact on execution, it is possible it will as there will be less time between each command as it is being sent. 

Another approach would be to mock `DoOnTime` for these tests (all devices?), so that its execution isnt relying on mocking time, and could instead be manually instructed to execute up to a certain point and allow the test to know that the queue will have been drained.

PR #244, opened 2023-02-22
feat(vmix): add support for starting and stopping VB.NET scripts

alvancamp:

Feature

vMix scripts cannot be started or stopped with TSR.

vMix scripts can be invoked and cancelled now:

```ts
export const input: TSRInput = {
	timeline: [
		literal<TSRTimelineObj<TimelineContentVMixScript>>({
			id: 'script0',
			enable: {
				start: Date.now(),
				duration: 20 * 1000,
			},
			layer: 'vmixLayer0',
			content: {
				deviceType: DeviceType.VMIX,
				type: TimelineContentTypeVMix.SCRIPT,
				name: 'myscript',
			},
		}),
	],
}
```

* **Other information**:

Tested via quick-tsr by writing a short VB.NET script that just has two console writes with a sleep between them:
```vb
Console.WriteLine("I sleep...")
Sleep(5000)
Console.WriteLine("I wake!!!")
```

### DIFF
--- a/packages/timeline-state-resolver/src/integrations/vmix/index.ts
+++ b/packages/timeline-state-resolver/src/integrations/vmix/index.ts
@@ -197,6 +197,7 @@ export class VMixDevice extends DeviceWithState<VMixStateExtended, DeviceOptions
 				Fullscreen2: { source: 'Program' },
 			},
 			inputLayers: {},
+			desiredScripts: [],
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -1015,6 +1056,7 @@ export interface VMixStateExtended {
 		Fullscreen2: VMixOutput
 	}
 	inputLayers: { [key: string]: string }
+	desiredScripts: string[]
```
jstarpl, line 1059: Why is this property called `desiredScripts`? Isn't the entirety of the state - _the desired state_? I think it might create some confusion down the line. I'd call it `scripts` or `runningScripts`, since as far as TSR is concerned, these scripts are running and, for example, it won't start the script again, if it thinks it is running because one timelineObject is replaced by another, with the same `SCRIPT`.

jstarpl (approved):
I'm not sure about calling it `desiredScripts` in the state object. Otherwise it looks good.

PR #249, opened 2023-03-08
fix: properly parse multi-packet vMix TCP API responses

alvancamp:

Bug fix

TSR fails to parse vMix TCP API responses which span multiple packets, due to a miscalculation of the response **body** length versus the response **header** length.

When checking to see if a response **body** has fully come in, the response **header** length is subtracted from the expected length of the body.

* **Other information**:

Debugged and confirmed using `quick-tsr `🙆‍♂️


jstarpl (approved):
This will need to be backported, possibly all the way back to R47.

jstarpl:
I was also under the impression that the test cases covered the broken-up packet situation. Since apparently that's not the case, we need to improve the test case.

PR #252, opened 2023-03-15
Expose makeReady functionality through TSR Actions

baltedewit:
This PR exposes any functionality that was previously used through the makeReady functions into TSR actions for use by library consumers. In addition, this PR marks the makeReady functionality as deprecated and to be removed in a later release.


packages/timeline-state-resolver/src/integrations/pharos/$schemas/actions.json
```diff
@@ -0,0 +1,10 @@
+{
+	"$schema": "../../../$schemas/action-schema.json",
+	"actions": [
+		{
+			"id": "clearQueue",
```
Julusian, line 5: Something else to question for cleanup, why do we have to explicitly do this clear? Shouldnt that be handled for us by the conductor due to the changing timeline?

Julusian (approved):
There should definitely be some cleanup following this, I am unsure if that makes sense to do in this PR, or if it should be done later as it will be a breaking change.
As part of this cleanup we should make sure to remove the 'make ready commands' from the httpsend device (and I think somewhere else), as I think they will no longer be used 

It would be nice to see some of these new methods being made private, to make it clear that they wont ever be called from outside of the class.

packages/timeline-state-resolver/src/integrations/atem/$schemas/actions.json
```diff
@@ -5,6 +5,11 @@
 			"id": "resync",
 			"name": "Resync Atem",
 			"destructive": false
+		},
+		{
+			"id": "logNextState",
```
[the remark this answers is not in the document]
baltedewit: I don't think it is, and since it has to be called explicitly now i think I'll just remove it in this PR now that I'm making changes anyway

packages/timeline-state-resolver/src/integrations/pharos/$schemas/actions.json
baltedewit: I agree, I don't see why this was in makeReady in the first place. I didn't really look at what the individual makeReady functions did but I'm just going to remove this one.

nytamin: Agreed, this one could probably be removed 👍

PR #258, opened 2023-05-03
fix(vmix): change how commands are ordered to reduce flashes of content in PGM

alvancamp:

Arguably a bug fix, arguably a feature.

TSR orders commands sent to the vMix device in a way that can sometimes cause a flash of content in PGM due to on-air Inputs having their properties changed before the transition occurs.

Input-altering commands are split into two arrays: one that gets executed before the transition and one that gets executed after. At present, the only logic for what differentiates between the two is whether or not a given Input is currently on-air. If it's currently on-air, then changes to it are executed _after_ the transition commands, instead of before. In most cases, this prevents flashes of content in PGM.

* **Other information**:

Edge cases may exist, though given that the existing implementation already has easily-hittable edge cases, I think this is fine to merge as a first attempt that can be refined later as edge cases are discovered.

Example "before" footage, where the clip player is cleared by TSR before the transition happens, resulting in several frames of black being visible:
https://user-images.githubusercontent.com/873012/235977586-28d9cf0b-eddd-4b8a-a4fe-a2bfa0ab3d4d.mp4

Example "after" footage, where the clip player isn't cleared by TSR until after the transition happens (the cut in the footage that occurs just before the transition is a natural cut in the footage and has nothing to do with this PR, I just timed my transition poorly relative to that in-footage cut):
https://user-images.githubusercontent.com/873012/235977679-6a31dee9-e8dd-4379-b3b5-84bf286543e7.mp4


packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -1155,8 +1170,10 @@ export class VMixDevice extends DeviceWithState<VMixStateExtended, DeviceOptions
 	): Array<VMixStateCommandWithContext> {
 		let commands: Array<VMixStateCommandWithContext> = []
 
-		commands = commands.concat(this._resolveInputsState(oldVMixState, newVMixState))
+		const inputCommands = this._resolveInputsState(oldVMixState, newVMixState)
+		commands = commands.concat(inputCommands.preTransitionCommands)
 		commands = commands.concat(this._resolveMixState(oldVMixState, newVMixState))
+		commands = commands.concat(inputCommands.postTransitionCommands)
```
ianshade, line 1176: Shouldn't this appear after resolving overlays state? And shouldn't _isOnAir consider inputs on overlays? What about outputs? You probably wouldn't like an output that is currently fed a list input to show a couple of black frames before it cuts to the next input after the list gets wiped?

ianshade: While this is probably sufficient for your use case, the biggest drawback of this approach is that there's no common definition of _on-air_. An input may appear on an output or in a Mix, but not be considered on-air for numerous reasons. This may cause inconsistencies in preparing inputs just before being taken, and will still require the user to apply some kind of pre-roll.

jstarpl (approved):

PR #265, opened 2023-06-13
feat!: Use strings for DeviceType enum

thomasslee97:
This PR has been opened by SuperFly.tv on behalf of EVS Broadcast Equipment.

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug fix.

**What is the current behavior?** (You can also link to an open issue here)
TSR is unable to initialize devices because device types are being stored as strings whereas TSR is expecting numbers. This is due to the keys of JS objects being strings, whereas TSR is attempting to use numbers.

**What is the new behavior (if this is a feature change)?**
The `DeviceType` enum has been changed to use strings rather than numbers, with the addtional benefit that devices will now have unique types - reducing the chance of a clash when new integrations are added.

**Other information**:


nytamin (changes requested):
@thomasslee97 could you point out a place where the strings-vs-numbers is an issue?

Changing the deviceType is a _very_ breaking change for consumers of the TSR library, so unless we don't have any other option I really want to avoid this.

Julusian:
> @thomasslee97 could you point out a place where the strings-vs-numbers is an issue?

This was spawned from a UI bug in Sofie. It would be simple to work around this, but after a quick discussion noone else on the team had an objection to this being the solution.  
Admittedly a UI bug is not a great reason for a breaking change, but this is something we have wanted to do for a while anyway but lacked a reason to do it.

> Changing the deviceType is a very breaking change for consumers of the TSR library, so unless we don't have any other option I really want to avoid this.

I don't think this is any more breaking than https://github.com/nrkno/sofie-timeline-state-resolver/pull/237 (also in release50) which renamed types, added enums to mappings/timelineobjects where there were previously none and rearranged the properties inside of Mappings to make them suitable to json schemas.

For example, a mapping such as
```
const a: MappingCasparCG = {
	device: TSR.DeviceType.CASPARCG,
	deviceId: 'caspar01',
	lookahead: LookaheadMode.NONE,
	channel: 3,
	layer: 199,
}
```
with that change must now be
```
const a: MappingCasparCG = {
	device: TSR.DeviceType.CASPARCG,
	deviceId: deviceId01,
	lookahead: LookaheadMode.PRELOAD,
	options: {
		mappingType: TSR.MappingCasparCGType.Layer,
		channel: 3,
		layer: 199,
	},
}
```

Julusian (approved):

thomasslee97:
I should also note that this was causing TSR to be unable to create devices, as TSR is trying to match against numbers to find a device integration but it is instead receiving a string (prior to this fix). This fix means that TSR is both trying to match against a string and is receiving a string to match against.

jstarpl:
> @thomasslee97 could you point out a place where the strings-vs-numbers is an issue?
> 
> Changing the deviceType is a _very_ breaking change for consumers of the TSR library, so unless we don't have any other option I really want to avoid this.

Well, this is hardly the first or the last breaking change in TSR. We've already had quite a number of these, and this one doesn't even seem that big. The Release 50 branch contains a significant amount of changes anyway, the upgrade/migration path to the new Enum seems trivial and this solves a long-standing architectural problem, impacting both contributions and users (as the Sofie Core example demonstrates)

nytamin:
> > @thomasslee97 could you point out a place where the strings-vs-numbers is an issue?
> > Changing the deviceType is a _very_ breaking change for consumers of the TSR library, so unless we don't have any other option I really want to avoid this.
> 
> Well, this is hardly the first or the last breaking change in TSR. We've already had quite a number of these, and this one doesn't even seem that big. The Release 50 branch contains a significant amount of changes anyway, the upgrade/migration path to the new Enum seems trivial and this solves a long-standing architectural problem, impacting both contributions and users (as the Sofie Core example demonstrates).
> 
> If anything, since there already will have to be a breaking release, this seems like the right time to do it.

Agreed, now would probably be a good time to do it.. Let's remember to make a note of it in the changelog though, and possibly change the commit message to a "feat!:".

nytamin (approved):

PR #267, opened 2023-06-18
fix: Update changed typings

PeterC89:
Bug fix

QuickTSR fails to compile due to TypeScript errors

QuickTSR compiles

* **Other information**:


Julusian:
Thanks, it appears we didn't make quick-tsr be built except when explicitly used. It will now be built in CI to make this more noticable

PR #269, opened 2023-07-04
fix(httpSend): Use the same types for the sendCommand action as a timeline object

PeterC89:
Bug fix

httpSend SendCommand TSR action has a separate less well typed payload (and is missing the headers property)

SendCommand payload now inherits from the normal httpSend TSR type

### DIFF
--- a/packages/timeline-state-resolver-types/src/generated/httpSend.ts
+++ b/packages/timeline-state-resolver-types/src/generated/httpSend.ts
@@ -46,16 +46,7 @@ export enum TimelineContentTypeHTTPParamType {
 
 export type SomeMappingHttpSend = Record<string, never>
 
-export interface SendCommandPayload {
-	type: string
-	url: string
-	params: {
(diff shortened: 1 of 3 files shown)


Julusian:
I think we need to keep the full definition in the json schema, as we are building ui components based on that schema.

It was broken until just now, but in the system status page, these actions can be exposed:
image
and clicking on 'Send command' will show a modal prompting for the values:
image
With the fields driven by the json schema

For now this means that certain types have to be duplicated, until someone puts in the work to figure out how references can be utilised for these shared types.

PeterC89:
@Julusian I've made a tweak that restores the JSON schema but keeps it referencing the TypeScript types, any chance you could see if that works?
I'm currently struggling to get Meteor to play nice on my machine for local dev... (Probably need to give up and just use WSL)

Julusian:
This is published in 9.0.0-nightly-release50-20230705-084118-3385217af.0

PR #271, opened 2023-07-12
chore: convert http device & cache dns reqs

baltedewit:

No DNS cache

Requests are cached, respecting the TTL of the records.

* **Other information**:

Also updates the HTTP send device to the new format

### DIFF
--- a/packages/timeline-state-resolver-types/src/generated/httpSend.ts
+++ b/packages/timeline-state-resolver-types/src/generated/httpSend.ts
@@ -7,41 +7,9 @@
 
 export interface HTTPSendOptions {
 	/**
-	 * Whether a makeReady should be treated as a reset of the device. It should be assumed clean, with the queue discarded, and state reapplied from empty
-	 */
-	makeReadyDoesReset?: boolean
-	/**
 	 * Minimum time in ms before a command is resent, set to <= 0 or undefined to disable
 	 */
 	resendTime?: number
-	makeReadyCommands?: HTTPSendCommandContent[]
-}
-export interface HTTPSendCommandContent {
-	type: TimelineContentTypeHTTP
-	url: string
-	params: {
-		[k: string]: unknown
-	}
-	paramsType?: TimelineContentTypeHTTPParamType
-	headers?: {
-		[k: string]: string
-	}
-	temporalPriority?: number
-	/**
-	 * Commands in the same queue will be sent in order (will wait for the previous to finish before sending next
-	 */
-	queueId?: string
-}
-
-export enum TimelineContentTypeHTTP {
-	GET = 'get',
-	POST = 'post',
-	PUT = 'put',
-	DELETE = 'delete'
-}
-export enum TimelineContentTypeHTTPParamType {
-	JSON = 'json',
-	FORM = 'form'
 }
 
 export type SomeMappingHttpSend = Record<string, never>
--- a/packages/timeline-state-resolver/src/integrations/httpSend/index.ts
+++ b/packages/timeline-state-resolver/src/integrations/httpSend/index.ts
@@ -1,154 +1,76 @@
-import * as _ from 'underscore'
-import { DeviceWithState, CommandWithContext, DeviceStatus, StatusCode } from './../../devices/device'
+import { EventEmitter } from 'eventemitter3'
+import { CommandWithContext, Device, DeviceEvents } from '../../service/device'
 import {
-	DeviceType,
-	HTTPSendOptions,
-	HTTPSendCommandContent,
-	DeviceOptionsHTTPSend,
-	Mappings,
-	Timeline,
-	TSRTimelineContent,
-	TimelineContentTypeHTTP,
-	TimelineContentTypeHTTPParamType,
 	ActionExecutionResult,
 	ActionExecutionResultCode,
+	DeviceOptionsHTTPSend,
+	DeviceStatus,
+	HTTPSendCommandContent,
+	HTTPSendOptions,
 	HttpSendActions,
 	SendCommandPayload,
+	StatusCode,
+	TSRTimelineContent,
+	Timeline,
+	TimelineContentTypeHTTP,
+	TimelineContentTypeHTTPParamType,
 } from 'timeline-state-resolver-types'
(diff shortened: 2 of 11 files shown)


packages/timeline-state-resolver/src/service/remoteDeviceInstance.ts
```diff
@@ -125,11 +125,11 @@ export class RemoteDeviceInstance<
 		getCurrentTime: () => number,
 		threadConfig?: ThreadedClassConfig
 	): Promise<RemoteDeviceInstance<TOptions>> {
-		if (process.env.JEST_WORKER_ID !== undefined && threadConfig && threadConfig.disableMultithreading) {
-			// running in Jest test environment.
-			// hack: we need to work around the mangling performed by threadedClass, as getCurrentTime needs to not return a promise
-			getCurrentTime = { inner: getCurrentTime } as any
-		}
+		// if (process.env.JEST_WORKER_ID !== undefined && threadConfig && threadConfig.disableMultithreading) {
```
Julusian, line 128: If this isn't needed, perhaps teh code should be removed?

packages/timeline-state-resolver-types/src/generated/httpSend.ts
```diff
@@ -6,42 +6,10 @@
...
-	 * Whether a makeReady should be treated as a reset of the device. It should be assumed clean, with the queue discarded, and state reapplied from empty
-	 */
-	makeReadyDoesReset?: boolean
 	/**
 	 * Minimum time in ms before a command is resent, set to <= 0 or undefined to disable
 	 */
 	resendTime?: number
-	makeReadyCommands?: HTTPSendCommandContent[]
-}
-export interface HTTPSendCommandContent {
```
Julusian, line 19: This change might conflict slightly with https://github.com/nrkno/sofie-timeline-state-resolver/pull/269/files, it would be good to make sure all changes from releaase50 are in the release51 branch and rebase this on top to make sure it does all work.

packages/timeline-state-resolver/src/integrations/httpSend/index.ts
```diff
@@ -366,10 +246,18 @@ export class HTTPSendDevice extends DeviceWithState<HTTPSendState, DeviceOptions
 					'EAI_AGAIN',
 				]
 
-				if (retryCodes.includes(err.code) && this._resendTime) {
-					const timeLeft = Math.max(this._resendTime - (Date.now() - t), 0)
-					await new Promise<void>((resolve) => setTimeout(() => resolve(), timeLeft))
-					this._defaultCommandReceiver(_time, cmd, context, timelineObjId, layer).catch(() => null) // errors will be emitted
+				if (retryCodes.includes(err.code) && this.options?.resendTime) {
+					const timeLeft = Math.max(this.options.resendTime - (Date.now() - t), 0)
+					setTimeout(() => {
```
Julusian, line 251: This probably isn't a new bug, but this retry will keep going after the device is disposed.
This could be solved by a simple check eitehr inside this `setTimeout`, or at the start of `sendCommand`, to return early if the device has been disposed

packages/timeline-state-resolver/src/integrations/httpSend/index.ts
```diff
@@ -1,154 +1,76 @@
...
-			if (this._makeReadyDoesReset) {
-				await this.resync()
-			}
 
-			for (const cmd of this._makeReadyCommands || []) {
-				await this.sendCommand(cmd)
-			}
-		}
+	actions: Record<string, (id: HttpSendActions, payload?: Record<string, any>) => Promise<ActionExecutionResult>> = {
+		[HttpSendActions.SendCommand]: async (_id: HttpSendActions.SendCommand, payload?: SendCommandPayload) =>
```
Julusian, line 64: Should the `resync` action also be implemented here, or is that handled elsewhere?

packages/timeline-state-resolver/src/integrations/httpSend/index.ts
```diff
@@ -366,10 +246,18 @@ export class HTTPSendDevice extends DeviceWithState<HTTPSendState, DeviceOptions
...
 				]
 
-				if (retryCodes.includes(err.code) && this._resendTime) {
-					const timeLeft = Math.max(this._resendTime - (Date.now() - t), 0)
-					await new Promise<void>((resolve) => setTimeout(() => resolve(), timeLeft))
-					this._defaultCommandReceiver(_time, cmd, context, timelineObjId, layer).catch(() => null) // errors will be emitted
+				if (retryCodes.includes(err.code) && this.options?.resendTime) {
+					const timeLeft = Math.max(this.options.resendTime - (Date.now() - t), 0)
+					setTimeout(() => {
+						this.sendCommand({
```
Julusian, line 252: Because this is run inside a `setTimeout`, `sendCommand` will return before the retry attempt is performed. Is this desired?
This does look to match what was done previously

packages/timeline-state-resolver/src/integrations/httpSend/index.ts
```diff
@@ -174,157 +96,112 @@ export class HTTPSendDevice extends DeviceWithState<HTTPSendState, DeviceOptions
 			}
 		}
 
-		await this._commandReceiver(time, cmd as HTTPSendCommandContent, 'makeReady', '')
+		await this.sendCommand({
```
Julusian, line 99: Not a new issue, but if the http request fails, this will result in a rejected promise for `sendManualCommand`, instead of an `ActionExecutionResult`. Should it?

Julusian (approved):

packages/timeline-state-resolver/src/integrations/httpSend/index.ts
baltedewit: Yeah, there is also a problem where commands sent through tsr actions will be resent forever. I think I'll add something quick n dirty for both edge cases.

packages/timeline-state-resolver/src/service/remoteDeviceInstance.ts
baltedewit: Yeah, probably. I just don't understand why it is not necessary now when it was before.

PR #272, opened 2023-07-27
fix(vmix): fix scenario where the media load retry system would load clips into playlists twice

alvancamp:

Bug fix and feature, kind of.

Under common conditions, the existing system to retry loading missing media will add that media to List Inputs (aka playlists) exactly twice, resulting in bad behavior in playout.

This PR completely changes the mechanism used for this system. Instead of detecting failed commands, we instead poll the state of vMix and use that to determine how different the real state of vMix is from our expected state. This difference is then used to generate new commands, same as any other state update.

* **Other information**:

As a side effect of this, more things than just media retrying will be present. Other commands that have failed unexpectedly may also get retried. This could have unforeseen consequences. This approach must be thought about carefully and alternatives or modifications should be considered before merging.

I'll be honest, some of this feels gross to me and I want there to be a better way, but maybe there isn't one (because vMix's API is... limiting.)

### DIFF
--- a/packages/timeline-state-resolver-types/src/generated/vmix.ts
+++ b/packages/timeline-state-resolver-types/src/generated/vmix.ts
@@ -9,9 +9,9 @@ export interface VMixOptions {
 	host: string
 	port: number
 	/**
-	 * How often, in milliseconds, to try re-sending certain failed commands. Defaults to 10 seconds. Set to 0 to disable. As of June 9th, 2023, this only works for media loading commands.
+	 * How often, in milliseconds, for when we should poll vMix to query its actual state. Used to know when to re-send certain failed commands.
 	 */
-	retryInterval?: number
+	pollInterval?: number
 }
 
 export interface MappingVmixProgram {
--- a/packages/timeline-state-resolver/src/integrations/vmix/$schemas/options.json
+++ b/packages/timeline-state-resolver/src/integrations/vmix/$schemas/options.json
@@ -11,10 +11,10 @@
 			"type": "integer",
 			"ui:title": "Port"
 		},
-		"retryInterval": {
+		"pollInterval": {
 			"type": "number",
 			"ui:title": "Retry Interval",
-			"description": "How often, in milliseconds, to try re-sending certain failed commands. Defaults to 10 seconds. Set to 0 to disable. As of June 9th, 2023, this only works for media loading commands."
+			"description": "How often, in milliseconds, for when we should poll vMix to query its actual state. Used to know when to re-send certain failed commands."
 		}
 	},
 	"required": ["host", "port"],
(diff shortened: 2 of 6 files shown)


jstarpl:
> As a side effect of this, more things than just media retrying will be present. Other commands that have failed unexpectedly may also get retried. This could have unforeseen consequences. This approach must be thought about carefully and alternatives or modifications should be considered before merging.

Quite honestly, I don't think that this is that big of a deal. The same thing can happen should a device handler or PoGw restart or should a RouteSet/mapping change. We expect the device to be in the defined state at all times and we expect it to be able to get it into the desired state at any time. Conversely, having a device in an _undefined_ state, can also lead to unforseen consequences - with the state being refreshed and commands being sent to get it back to a _defined_ state - the worst case scenario is that we will do exactly what we should, even if the user sort-of accepted living with a command failure somewhere. As long as whatever mitigating actions they take stay within Sofie (so modifying the desired state, as opposed to operating on an undefined state), everything should be fine?

jstarpl:
> The bits that bug me the most are [#272 (files)](https://github.com/nrkno/sofie-timeline-state-resolver/pull/272/files#diff-57b14dabbb84e4b71ce374610e7dc0bb2d10a844bc670d1b60e781d0a2f0d786R317) and [#272 (files)](https://github.com/nrkno/sofie-timeline-state-resolver/pull/272/files#diff-2bed5085ac736505c2db24ac46af3b4e6682c6f54857a635b6f3837bd8eeb2bfR162-R174). These are the code blocks responsible for knowing which properties to "carry over" from the old state, because they cannot be computed from vMix's API. These bits bug me because they seem fragile, easy to change one without the other, etc. I couldn't think of anything better and it's possible that nothing better exists.

I see your point - what about somehow splitting this object into two? One that has the "receivedPartialState" and then another, that has the "inferredPartialState"? This way the ownership is clear, and you wouldn't have to copy anything over. I guess it will still be brittle (there being two states instead of one and them needing to be logically "merged" when used), but maybe that's an option?

> [#272 (files)](https://github.com/nrkno/sofie-timeline-state-resolver/pull/272/files#diff-2bed5085ac736505c2db24ac46af3b4e6682c6f54857a635b6f3837bd8eeb2bfR162-R174) also worries me because it seems like it might get users into a situation where they want media to be retried, but they're still actively producing and sending commands, which will indefinitely push the reloading away into the future, 10 seconds at a time.

Okay, I guess this is a very reasonable worry. 10 seconds is a lot of time and I can easily see something happening every <= 10, pushing this refresh away. Maybe this needs to be limited? Allow the refresh to be re-scheduled, but by no more than X? Or have it on a dynamic scale, where every reschedule halves the timespan, and the timespan is reset once the refresh happens?

thomasslee97:
> I'm happy with this now. Thanks @jstarpl for the inspiration on how to make the code more robust such that it is difficult to update things in one place but not the other. The types feel safer as a result.
> 
> If possible, I'd love a quick glance from @thomasslee97 because I'm told that this kind of "real-world actual device state" system is something he has knowledge and perhaps opinions on.

The work I've been doing has only been looking at the initial connection, we've not started any work on tracking against real-world state during the course of a show. But even if we had, we have more useful error types to work with and a more granular API to fetch state from.

baltedewit (approved):

jesperstarkar:
Continues in https://github.com/nrkno/sofie-timeline-state-resolver/pull/277

PR #273, opened 2023-08-01
chore: refactor shotoku device

mint-dewit:

* **Other information**:

### DIFF
--- a/packages/timeline-state-resolver/src/integrations/shotoku/index.ts
+++ b/packages/timeline-state-resolver/src/integrations/shotoku/index.ts
@@ -2,7 +2,6 @@ import {
 	ActionExecutionResult,
 	DeviceStatus,
 	DeviceType,
-	OSCOptions,
 	StatusCode,
 	Timeline,
 	TimelineContentShotokuSequence,
@@ -11,16 +10,12 @@ import {
 	TimelineContentTypeShotoku,
 	ShotokuTransitionType,
 	ShotokuOptions,
-	DeviceOptionsShotoku,
 } from 'timeline-state-resolver-types'
 import { Device, DeviceEvents } from '../../service/device'
 
 import _ = require('underscore')
 import EventEmitter = require('eventemitter3')
 import { ShotokuAPI, ShotokuCommand, ShotokuCommandType } from './connection'
-
-export type ShotokuDeviceOptions = ShotokuOptions
-export type DeviceOptionsShotokuInternal = DeviceOptionsShotoku
 
 export interface ShotokuDeviceState {
 	shots: Record<string, ShotokuCommandContent & { fromTlObject: string }>
@@ -29,9 +24,6 @@ export interface ShotokuDeviceState {
 interface ShotokuSequence {
 	fromTlObject: string
 	shots: TimelineContentShotokuSequence['shots']
-}
-interface ShotokuDeviceStateContent extends ShotokuCommandContent {
-	fromTlObject: string
 }
 
 export interface ShotokuCommandWithContext {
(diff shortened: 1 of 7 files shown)


packages/timeline-state-resolver/src/integrations/shotoku/index.ts
```diff
@@ -1,154 +1,64 @@
...
+	context: string
+	tlObjId: string
+}
 
-		const convertTrace = startTrace(`device:convertState`, { deviceId: this.deviceId })
-		const newShotokuState = this.convertStateToShotokuShots(newState)
-		this.emit('timeTrace', endTrace(convertTrace))
+export class ShotokuDevice
+	extends EventEmitter<DeviceEvents>
+	implements Device<OSCOptions, ShotokuDeviceState, ShotokuCommandWithContext>
```
Julusian, line 45:
```suggestion
	implements Device<ShotokuOptions, ShotokuDeviceState, ShotokuCommandWithContext>
```

packages/timeline-state-resolver/src/integrations/shotoku/index.ts
```diff
@@ -178,42 +88,13 @@ export class ShotokuDevice extends DeviceWithState<ShotokuDeviceState, DeviceOpt
...
-	 * @param newState The desired state of the device
-	 */
-	private _diffStates(oldState: ShotokuDeviceState, newState: ShotokuDeviceState): Array<Command> {
+	diffStates(oldState: ShotokuDeviceState | undefined, newState: ShotokuDeviceState): Array<ShotokuCommandWithContext> {
 		// unfortunately we don't know what shots belong to what camera, so we can't do anything smart
 
-		const commands: Array<Command> = []
+		const commands: Array<ShotokuCommandWithContext> = []
 
 		_.each(newState.shots, (newCommandContent: ShotokuDeviceStateContent, index: string) => {
```
Julusian, line 96: While you're here, it would be nice to remove the use of underscore. Not essential though

packages/timeline-state-resolver/src/integrations/shotoku/index.ts
```diff
@@ -260,30 +141,32 @@ export class ShotokuDevice extends DeviceWithState<ShotokuDeviceState, DeviceOpt
...
-			timelineObjId: timelineObjId,
-		}
-		this.emitDebug(cwc)
+	async sendCommand({ command, context, tlObjId }: ShotokuCommandWithContext): Promise<any> {
+		this.emit('debug', { command, context, tlObjId })
 
 		try {
 			if (this._shotoku.connected) {
-				this._shotoku.executeCommand(cmd).catch((e) => {
+				this._shotoku.executeCommand(command).catch((e) => {
```
Julusian, line 149: This needs to be awaited.
The rethrown error will uncaught, and will likely crash the thread

Julusian (approved):
You haven't added the device to the `DevicesDict`, so it won't be loaded (I've got a commit which will enforce this through typescript coming)

packages/timeline-state-resolver/src/integrations/shotoku/index.ts
```diff
@@ -1,154 +1,64 @@
...
-) => Promise<any>
-interface Command {
-	command: ShotokuCommand
-	context: CommandContext
-	timelineObjId: string
-}
-type CommandContext = string
-type ShotokuDeviceState = {
+export type ShotokuDeviceOptions = ShotokuOptions
+export type DeviceOptionsShotokuInternal = DeviceOptionsShotoku
```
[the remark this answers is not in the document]
mint-dewit: Yeah, I think they were used in the conductor but I'm going to go ahead and remove the usage of them there. They're already exported from the types package so no real need to export more things.

packages/timeline-state-resolver/src/integrations/shotoku/index.ts
```diff
@@ -1,154 +1,64 @@
...
-		this.emit('timeTrace', endTrace(diffTrace))
+	async init(options: ShotokuDeviceOptions): Promise<boolean> {
+		this._shotoku = new ShotokuAPI()
+		this._shotoku.on('error', (info, e) => this.emit('error', e, info))
 
-		// clear any queued commands later than this time:
-		this._doOnTime.clearQueueNowAndAfter(previousStateTime)
-		// add the new commands to the queue:
-		this._addToQueue(commandsToAchieveState, newState.time)
+		await this._shotoku.connect(options.host, options.port)
```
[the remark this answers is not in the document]
mint-dewit: Good spot. I don't think anything has changed in regards to how this is handled, but it's much better to write the devices in such a way that they will retry the connection internally. Initialisation should then only fail if some precondition is missing.

After all, a device can just as well disconnect during regular operations and we don't want to restart the device for that either.

PR #275, opened 2023-08-10
feat(sisyfos): Send fadeTime with faderLevel if specified

PeterC89:
Feature (sisyfos)

Sisyfos cannot change fade time if the fader `isPgm` value isn't changed, so level changes cannot have adjustable fade time.

In conjunction with https://github.com/tv2/sisyfos-audio-controller/pull/271 the `fadeTime` property is sent with level changes.

* **Other information**:
Tested against an unmodified version of Sisyfos (4.18.0) and appears to have no ill effects so should be fine to be in TSR until the Sisyfos PR is merged.


mint-dewit (approved):

jstarpl:
@PeterC89 Technically, the `release50` is feature-frozen. How big of a problem would it be for you if we pushed this into `release51` instead?

PeterC89:
@jstarpl Not a problem at all, I can always build playout gateway manually if I need it urgently

PR #276, opened 2023-08-14
chore: refactor abstract device SOFIE-2479

Julusian:


Julusian:
I am ignoring the one failing unit test, as it is reproducible in release50 and will need fixing there

packages/timeline-state-resolver/src/integrations/abstract/index.ts
```diff
@@ -1,245 +1,128 @@
...
-	async terminate() {
-		this._doOnTime.dispose()
-	}
-	get canConnect(): boolean {
-		return false
+	async terminate(): Promise<void> {
+		// Noop
 	}
+
 	get connected(): boolean {
```
mint-dewit, line 49: This might be expressed more simply as a public property.

packages/timeline-state-resolver/src/integrations/abstract/__tests__/abstract.spec.ts
```diff
@@ -1,185 +1,172 @@
...
-			getCurrentTime: mockTime.getCurrentTime,
-		})
-		await myConductor.init()
-		await myConductor.addDevice('myAbstract', {
-			type: DeviceType.ABSTRACT,
-			options: {},
-			commandReceiver: commandReceiver0,
-		})
-		await mockTime.advanceTimeToTicks(10100)
+	test('Device setup', async () => {
```
mint-dewit, line 21: Perhaps this could have another test describing no commands should be sent, but it's fine without imo.

mint-dewit (approved):
This looks good, I really like how we're mocking the device instances in the unit tests for the conductor now.

I think I should probably also add unit tests for the DeviceInstance classes. Fix for the one unit test is in the pipeline :)

packages/timeline-state-resolver/src/service/devices.ts
```diff
@@ -13,13 +14,20 @@ export interface DeviceEntry {
...
 	| DeviceType.HTTPSEND
 	| DeviceType.HTTPWATCHER
 	| DeviceType.OSC
 	| DeviceType.SHOTOKU
 
 // TODO - move all device implementations here and remove the old Device classes
 export const DevicesDict: Record<ImplementedServiceDeviceTypes, DeviceEntry> = {
+	[DeviceType.ABSTRACT]: {
+		deviceClass: AbstractDevice,
+		canConnect: true,
```
mint-dewit, line 27: I think canConnect should be false

PR #277, opened 2023-08-15
Feat/vmix media reload retrying

jesperstarkar:

Extending https://github.com/nrkno/sofie-timeline-state-resolver/pull/272

* **Other information**:

### DIFF
--- a/packages/timeline-state-resolver/src/integrations/vmix/$schemas/options.json
+++ b/packages/timeline-state-resolver/src/integrations/vmix/$schemas/options.json
@@ -13,7 +13,7 @@
 		},
 		"pollInterval": {
 			"type": "number",
-			"ui:title": "Retry Interval",
+			"ui:title": "Poll Interval",
 			"description": "How often, in milliseconds, for when we should poll vMix to query its actual state. Used to know when to re-send certain failed commands."
 		}
 	},
--- a/packages/timeline-state-resolver/src/integrations/vmix/$schemas/options.json
+++ b/packages/timeline-state-resolver/src/integrations/vmix/$schemas/options.json
@@ -14,7 +14,7 @@
 		"pollInterval": {
 			"type": "number",
 			"ui:title": "Poll Interval",
-			"description": "How often, in milliseconds, for when we should poll vMix to query its actual state. Used to know when to re-send certain failed commands."
+			"description": "How often, in milliseconds, for when we should poll vMix to query its actual state. Used to know when to re-send certain failed commands. Values <= 0 disables the polling. Defaults to 10000ms."
 		}
 	},
 	"required": ["host", "port"],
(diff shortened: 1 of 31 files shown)


jesperstarkar:
Pending testing

packages/timeline-state-resolver/src/integrations/vmix/$schemas/options.json
```diff
@@ -11,10 +11,10 @@
 			"type": "integer",
 			"ui:title": "Port"
 		},
-		"retryInterval": {
+		"pollInterval": {
```
nytamin, line 14: Technically this is a breaking change, I'm okay with it though.

packages/timeline-state-resolver/src/integrations/vmix/__tests__/vmix.spec.ts
```diff
@@ -211,6 +212,7 @@ describe('vMix', () => {
 
 		// Time to stop playing
 		await mockTime.advanceTimeToTicks(16050)
+		console.log(JSON.stringify(commandReceiver0.mock.calls, undefined, 4))
```
nytamin, line 215: remove this log line

packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -137,31 +131,25 @@ export class VMixDevice extends DeviceWithState<VMixStateExtended, DeviceOptions
...
 		})
 		// this._vmix.on('debug', (...args) => this.emitDebug(...args))
 
 		this._vmix.connect()
 
-		if (typeof options.retryInterval !== 'number' || options.retryInterval > 0) {
-			this._retryTime = options.retryInterval ?? DEFAULT_MEDIA_RETRY_INTERVAL
-			this._retryTimeout = setTimeout(() => this._assertIntendedState(), this._retryTime)
+		if (typeof options.pollInterval !== 'number' || options.pollInterval > 0) {
+			this._pollTime = options.pollInterval ?? DEFAULT_VMIX_POLL_INTERVAL
```
nytamin, line 146: This logic is pretty hard to follow.

```suggestion
		this._pollTime = options.pollInterval ?? DEFAULT_VMIX_POLL_INTERVAL // options.pollInterval  === 0 disables the polling
		if (this._pollTime  > 0 ) {
			
```

nytamin (approved):

packages/timeline-state-resolver/src/integrations/vmix/index.ts
jesperstarkar: Didn't end up with the exact same implementation.

One side effect of the proposed implementation is that non-numeric numbers will disable all polling, same as <= 0 nums.

If we want this case to fall to the default we need to unwrap this into two IFs.

jesperstarkar: A side-effect of (both the renaming and this change) is that we now have two blocks/implementations of similar-but-different code for the same prop/mechanism between CasparCG and vMix.

nytamin (approved):

PR #279, opened 2023-08-15
chore: refactor http-watch device SOFIE-2482

Julusian:


packages/timeline-state-resolver/src/service/device.ts
```diff
@@ -63,7 +63,7 @@ export interface BaseDeviceAPI<DeviceState, Command extends CommandWithContext>
 	 */
 	diffStates(oldState: DeviceState | undefined, newState: DeviceState, mappings: Mappings): Array<Command>
 	/** This method will take a command and send it to the device */
-	sendCommand(command: Command): Promise<any>
+	sendCommand(command: Command): Promise<void>
```
mint-dewit, line 66: `sendCommand` returns the result of the command in my design which is to be used in a generic handler for retries, media status and partial control. That means at some point I need these values to be non-void. Although using any is also not ideal here.

packages/timeline-state-resolver/src/service/device.ts
```diff
@@ -74,7 +74,7 @@ export interface DeviceEvents {
 
 	debugState: [state: object]
 	/** The connection status has changed */
-	connectionChanged: [status: DeviceStatus]
+	connectionChanged: [status: Omit<DeviceStatus, 'active'>]
```
mint-dewit, line 77: At this point we should probably have `DeviceStatus` extend another interface

packages/timeline-state-resolver/src/integrations/httpWatcher/index.ts
```diff
@@ -138,16 +124,20 @@ export class HTTPWatcherDevice extends Device<DeviceOptionsHTTPWatcherInternal>
...
-	get canConnect(): boolean {
-		return false
-	}
 	get connected(): boolean {
 		return false
 	}
-	get deviceType() {
-		return DeviceType.HTTPWATCHER
+
+	convertTimelineStateToDeviceState(_state: Timeline.TimelineState<TSRTimelineContent>): HTTPWatcherDeviceState {
```
mint-dewit, line 131: No need to have these arguments if we aren't using them anyway and aren't likely to be later.

packages/timeline-state-resolver/src/integrations/httpWatcher/index.ts
```diff
@@ -26,38 +33,8 @@ export class HTTPWatcherDevice extends Device<DeviceOptionsHTTPWatcherInternal>
...
-
-		this.expectedHttpResponse = Number(opts?.expectedHttpResponse) || undefined
-		this.headers = opts?.headers
-		this.keyword = opts?.keyword
-		this.intervalTime = Math.max(Number(opts?.interval) || 1000, 1000)
-		this.uri = opts?.uri
-	}
 
-	onInterval() {
+	#onInterval() {
```
mint-dewit, line 37: I am personally not a fan of using these private modifiers because they are so inconsistent with typescript. I wish typescript would still let you type them as private explicitly because I get a stroke reading this. Not a real issue with using them here, I just wish the world was a less triggering place.

packages/timeline-state-resolver/src/integrations/httpWatcher/index.ts
```diff
@@ -1,22 +1,29 @@
...
+	StatusCode,
+	DeviceStatus,
 } from 'timeline-state-resolver-types'
 import got, { Headers, Response } from 'got'
+import EventEmitter = require('eventemitter3')
+import { CommandWithContext, Device, DeviceEvents } from '../../service/device'
 
-export type DeviceOptionsHTTPWatcherInternal = DeviceOptionsHTTPWatcher
+type HTTPWatcherDeviceState = Record<string, never>
+type HTTPWatcherCommandWithContext = CommandWithContext
```
mint-dewit, line 15: Consider if this alias is necessary.

mint-dewit (approved):
I've been thinking about moving away from the event-driven model here and move towards providing the integrations with a context object instead. I feel like it's a bit more in-line with the relationship between the DeviceInstanceWrapper and the Integration implementation than how we use the events now. At the same time it's good to limit the interactions from implementation to wrapper to a minimum. Something to think about, but not for this PR.

I agree with moving the terminate return type to void, there's really not anything we can do if it doesn't return true anyway

packages/timeline-state-resolver/src/service/device.ts
Julusian: I don't have a problem with them being non-void, but `any` is a bit of a smell and was unnecessary based on usage. 
So change it when you get to that retry handler, until then `void` is more descriptive of the current intention

packages/timeline-state-resolver/src/service/device.ts
Julusian: I did think that, but didn't want to figure that out as part of this

packages/timeline-state-resolver/src/integrations/httpWatcher/index.ts
Julusian: I've changed them back. I do quite like them, but as I don't do much in this repository its better to stick to the trend of the other devices.

PR #270, opened 2023-07-07
Feat: Timeline v9 (SOFIE-2514)

nytamin:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

### Updates TSR to use the new Timeline version 9

This is a Big breaking change (see PR: https://github.com/SuperFlyTV/supertimeline/pull/95 )
which requires some type-changes in TSR as well as modified resolving-logic.

### Changes to the resolve-loop in conductor

Added `this._resolved.validTo` property, to better track when it is time to re-resolve the timeline.

Note: This PR Is pending on these PRs:

* https://github.com/SuperFlyTV/supertimeline/pull/95
* https://github.com/nrkno/sofie-timeline-state-resolver/pull/280

PR #280, opened 2023-08-18
Fix: vmix preloading

nytamin:
Changes in how vmix-preloading works

The vmix-preloading of media onto inputs is done on a separate command-queue (executed on `currentTime`) but generated from the stored `state`'s (tied to `resolveTime`).

This can cause issues because the sending of those commands aren't done in sync with the state-tracker, so multiple commands for the same thing might be generated. (I excperienced this issue in PR #270 ).

I moved away the handling of pre-loading command into a separate class that handles those commands only and tracks the actual state of those. So these commands are now sent asynchronously from the main command-queue, but that should be OK since we also tracks a separate state for them.

* **Other information**:


packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -1454,3 +1432,127 @@ export interface VMixAudioChannel {
 	meterF2: number
 	headphonesVolume: number
 }
+
+/**
+ * Handles pre-loading of inputs.
+ */
+class VMixInputHandler extends EventEmitter {
+	// how long to let an input linger before removing it
+	private TTL = 1000
```
ianshade, line 1441: This is already an improvement, but since we have some logic in `_resolveInputsState` that is able to reorder commands based on whether the input is (was?) on air or not, perhaps we could also use that to delay the removal by at least the duration of the transition. That's for another PR of course

packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -1454,3 +1432,127 @@ export interface VMixAudioChannel {
...
+			if (input.removeTime) {
+				nextTimeout = Math.min(nextTimeout, input.removeTime)
+			}
+		}
+
+		this._vmix.addToQueue(commands, now)
+
+		const timeToNextTimeout = nextTimeout - now
+		if (timeToNextTimeout > 0 && timeToNextTimeout !== Infinity) {
+			this._updateTimeout = setTimeout(() => {
```
ianshade, line 1549: I'm a little confused as to what will be the value of `timeToNextTimeout` when an input gets removed. Say `now := 10000`, and we have two inputs, one to be removed at 10000, another at 10500. Won't `nextTimeout` be set to 10000 and `timeToNextTimeout` to 0, causing a timeout not to be set, resulting in the other input lingering until a third input is scheduled to be removed?

ianshade: Perhaps just a `continue` is needed after [this line](https://github.com/nrkno/sofie-timeline-state-resolver/pull/280/files/39d3ebad6f175fc710bda7a94af6862a6452df80#diff-2bed5085ac736505c2db24ac46af3b4e6682c6f54857a635b6f3837bd8eeb2bfR1515)

nytamin: Good catch! I'll push a fix with a `continue`.

mint-dewit:
@alvancamp I was hoping you could give some insight into that actually, since you've been working with the vmix integration a lot

packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -275,13 +280,14 @@ export class VMixDevice extends DeviceWithState<VMixStateExtended, DeviceOptions
...
 
-		const commandsToAchieveState: Array<any> = this._diffStates(oldState, newVMixState)
+		const commandsToAchieveState: Array<any> = this._diffStates(newState.time, oldState, newVMixState)
 
 		// clear any queued commands later than this time:
 		this._doOnTime.clearQueueNowAndAfter(previousStateTime)
 
 		// add the new commands to the queue:
-		this._addToQueue(commandsToAchieveState, newState.time)
+
```
mint-dewit, line 289: Unnecessary white space

packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -1454,3 +1432,131 @@ export interface VMixAudioChannel {
 	meterF2: number
 	headphonesVolume: number
 }
+
+/**
+ * Handles pre-loading of inputs.
+ */
+class VMixInputHandler extends EventEmitter {
```
mint-dewit, line 1439: This class should be moved to a separate file as this file is already _much_ too long

packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -1454,3 +1432,131 @@ export interface VMixAudioChannel {
...
+			/** Is set to true when the commands to add the input has been set. */
+			addedCommandSent: boolean
+			/** Is set to something when the input is scheduled to be removed. */
+			removeTime?: number
+		}
+	> = new Map()
+
+	private _updateTimeout: NodeJS.Timeout | undefined = undefined
+
+	constructor(private _vmix: VMixDevice) {
```
mint-dewit, line 1458: Instead of passing the entire class to a subclass you should just pass callbacks for adding commands and getting the time. That will also make it easier to add unit tests in the future and refactor the device integration.

packages/timeline-state-resolver/src/integrations/vmix/index.ts
```diff
@@ -587,7 +589,7 @@ export class VMixDevice extends DeviceWithState<VMixStateExtended, DeviceOptions
 		return this._doOnTime.getQueue()
 	}
 
-	private _addToQueue(commandsToAchieveState: Array<VMixStateCommandWithContext>, time: number) {
+	public addToQueue(commandsToAchieveState: Array<VMixStateCommandWithContext>, time: number) {
```
mint-dewit, line 592: We shouldn't carelessly add things to the public API of these devices. If using callbacks this isn't necessary.

mint-dewit (changes requested):

packages/timeline-state-resolver/src/integrations/vmix/index.ts
mint-dewit: Yes, but I still want to be careful with adding methods on these device classes as they are exposed to anyone consuming the TSR.

PR #281, opened 2023-08-24
fix: resync chef on reconnect (SOFIE-2544)

nytamin:
Functionality fix

If Sofie Chef goes offline and then comes back, there will be no resync of state, so Chef will continue to display the default url.

Upon a reconnect, TSR resynct the state and sends new commands.

* **Other information**:

This PR depends on https://github.com/nrkno/sofie-chef/pull/6

### DIFF
--- a/packages/timeline-state-resolver/src/integrations/sofieChef/index.ts
+++ b/packages/timeline-state-resolver/src/integrations/sofieChef/index.ts
@@ -413,7 +413,9 @@ export class SofieChefDevice extends DeviceWithState<SofieChefState, DeviceOptio
 
 				// Is it in the mappings?
 				// We should only touch stuff that is in our mappings:
-				const foundMapping = Object.values(newMappings).find((mapping) => mapping.options.windowId === windowId)
+				const foundMapping = Object.values<Mapping<SomeMappingSofieChef>>(newMappings).find(
+					(mapping) => mapping.options.windowId === windowId
+				)
 
 				if (foundMapping) {
 					commands.push({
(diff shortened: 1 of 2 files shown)


packages/timeline-state-resolver/src/integrations/sofieChef/index.ts
```diff
@@ -373,15 +410,24 @@ export class SofieChefDevice extends DeviceWithState<SofieChefState, DeviceOptio
...
-					timelineObjId: oldWindow.urlTimelineObjId,
-					content: {
-						msgId: 0, // set later
-						type: ReceiveWSMessageType.STOP,
-						windowId: windowId,
-					},
-				})
+
+				// Is it in the mappings?
+				// We should only touch stuff that is in our mappings:
```
mint-dewit, line 415: Actually thinking about this a second time and now I'm even doubting this... mappings can change when route sets change and we maybe want to remove the window?

packages/timeline-state-resolver/src/integrations/sofieChef/index.ts
```diff
@@ -145,15 +150,40 @@ export class SofieChefDevice extends DeviceWithState<SofieChefState, DeviceOptio
...
+					// Resync state:
+					await this.resyncState()
 				})
 				.catch(() => {
 					// Unable to reconnect, try again later:
 					this.tryReconnect()
 				})
 		}, RECONNECT_WAIT_TIME)
 	}
+	async resyncState() {
```
mint-dewit, line 165: I don't think this method needs to be public? If it needs to be accessed outside of the device class perhaps adding a tsr action would be better.

packages/timeline-state-resolver/src/integrations/sofieChef/index.ts
nytamin: Fair enough! I'll remove the check for now and we can revisit it later when we're explicitly looking into partial control.

PR #284, opened 2023-08-31
chore: refactor atem device to state handler SOFIE-2490

Julusian:

* **Other information**:


mint-dewit (approved):
I think this is a good time to remove the (already deprecated) makeReady action.

Should the `_initialized` property be kept around? Is it significantly different from not being connected?

I like the approach taken here with the unit tests. I think we should be a bit careful with testing external libraries unnecessarily though. i.e. most of what `diffStates` outputs comes from the atem-state library. Not saying it's too much here either though.

Julusian:
> I think this is a good time to remove the (already deprecated) makeReady action.

Sure

> Should the `_initialized` property be kept around? Is it significantly different from not being connected?

I am not sure. It has ended up solely being a guard for ensuring that `init` is called. But perhaps this shouldn't care about that and should assume that it will be? I suppose it is part of the contract of the `Device` interface that it will be called appropriately.

> I like the approach taken here with the unit tests. I think we should be a bit careful with testing external libraries unnecessarily though. i.e. most of what `diffStates` outputs comes from the atem-state library. Not saying it's too much here either though.

Yeah, I almost did mock `atem-state` for this file but didnt. I think there is value in some tests, to ensure that things do line up fully, but I suppose the `full flow` ones in `atem.spec.ts` cover that.   
With #289 I think this file could have more purpose, and could definitely benefit from mocking for some of the tests. That PR should really have some more added here, with the focus being checking the 'diff' parameter being passed to `atem-state`. I also think it would be appropriate to check the output (in less detail) to ensure that things did/didnt diff as intended. While that will be testing `atem-state`, it would also be verifying our interpretation of what should happen

PR #285, opened 2023-08-31
Sisyfos device would not garbage collect (Sofie 2576)

mint-dewit:
When a device is terminated it should be ready for garbage collection. A lot of devices don't seem to do this very well, this isn't a problem when ran in multi threaded mode because the thread can be restarted but in single threaded mode can cause issues.

In addition the sisyfos device used a fixed local port which is unnecessary and prevented us from running multiple connections (combined with the above this caused connection instability)

### DIFF
--- a/packages/timeline-state-resolver/src/integrations/casparCG/index.ts
+++ b/packages/timeline-state-resolver/src/integrations/casparCG/index.ts
@@ -164,11 +164,15 @@ export class CasparCGDevice extends DeviceWithState<State, DeviceOptionsCasparCG
 			if (!this._ccg) {
 				resolve(true)
 				return
-			}
-			this._ccg.disconnect()
-			this._ccg.onDisconnected = () => {
-				resolve(true)
+			} else if (!this._ccg.connected) {
+				this._ccg.onDisconnected = () => {
+					resolve(true)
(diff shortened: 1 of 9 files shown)


packages/timeline-state-resolver/src/integrations/casparCG/index.ts
```diff
@@ -168,6 +168,7 @@ export class CasparCGDevice extends DeviceWithState<State, DeviceOptionsCasparCG
 			this._ccg.disconnect()
 			this._ccg.onDisconnected = () => {
 				resolve(true)
+				this._ccg.removeAllListeners()
```
nytamin, line 171: I think that there is actually a potential problem here. If the caspar connection happens to be disconnected here, the onDisconnected won't be called (see https://github.com/SuperFlyTV/casparcg-connection/blob/5.0.0/src/CasparCG.ts#L2206). So we should perhaps do this here:
```

this._ccg.disconnect()
this._ccg.onDisconnected = () => {
  resolve(true)
  this._ccg.removeAllListeners()
}
if (!this._ccg.connected) this._ccg.onDisconnected()
```

nytamin (approved):

PR #286, opened 2023-09-04
feat: Dereference schemas

sbaudlr:
This PR has been opened by SuperFly.tv on behalf of EVS Broadcast Equipment (henceforth EVS).

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Feature.

**What is the current behavior?** (You can also link to an open issue here)

Schemas must be in a single file.

**What is the new behavior (if this is a feature change)?**

Options, Mappings, and Actions schemas can use `"$ref"` in order to be split into multiple files and re-use definitions for complicated schemas.

**Other information**:

The `schema-types` script will now look for schemas that contain `"$ref"` and resolve the references. The resulting schema will be written to `${schema}.deref.json`, e.g. `mappings.deref.json` for a `mappings.json` file containing references. If a schema has been created with references, then `packages/timeline-state-resolver/src/manifest.ts` needs to be updated to include the `${schema}.deref.json` file. When generating the types for a schema, the dereferenced schema will automatically be used if it exists.


mint-dewit (commented):
From some experimentation it seems the library we use to generate typings is already happily resolving references, so it seems that what is missing is actually only the inclusion of these in the `manifest.ts` file?

I wonder then if it makes sense to do an extra step of generation in a script vs doing this step at runtime. I can see some benefits to both but primarily doing it at runtime would ensure a dev can't forget to include the right manifest when switching to using refs.

Julusian:
I don't have a strong preference on which approach is taken, but I do think that generating like this has some nice benefits; ensuring that invalid refs can be detected at build rather than runtime and simplifying the startup, avoiding needing to do a load of iteration and synchronous file loading at every startup.

As an added bonus (which has no impact on us, or any project I am aware of yet), this way will play much nicer with webpacking backend code (perhaps to use with https://nodejs.org/api/single-executable-applications.html), as it doesnt rely on the filesystem at runtime

sbaudlr:
I think it's nice to have a single schema file available from TSR, for example for importing into documentation.

mint-dewit:
I don't really agree that any mentioned concerns so far are practical issues, but I can tell that there is a preference for single files either way.

If we go with creating a single file at build time, I think we should save the generated schemas in a subfolder of the `$schemas` folder then. We should just always output these, regardless whether or not refs have been used. The script to generate the tsr-types stuff can always use the generated `.ref.json` files which also simplifies that step.

I don't want the generation script to behave very differently when refs are used, that will just be confusing later on and comes with implicit rules.

An alternative to this could be to generate the `manifest.ts` file too, although that needs to be given more thought.

sbaudlr:
I've changed the script to generate dereferenced schemas for everything and to place those generated files in one folder. I had a go at generating the `manifest.ts` but didn't manage to get anything I feel happy with.

mint-dewit (approved):

Julusian (approved):

PR #287, opened 2023-09-08
Chore: Release51, minor improvements

nytamin:
This PR touches a few things that I came across. Consider them to be suggestions to be cherry-picked or merged as a whole.

See commit log below.


packages/timeline-state-resolver/src/service/__tests__/devices.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+describe('Ensure that all integrations have defined their actions', () => {
+	for (const [key, device] of Object.entries<DeviceEntry>(DevicesDict)) {
+		test(`Device ${key}`, () => {
+			const deviceManifest = manifest.subdevices[key]
+			expect(deviceManifest).toBeTruthy()
+
+			const deviceInstance = new device.deviceClass()
+			expect(deviceInstance).toBeTruthy()
+
+			if (deviceManifest.actions) {
```
Julusian, line 14: This could return early to avoid an extra level of indentation

Julusian (approved):

PR #289, opened 2023-09-13
feat: update atem-connection and atem-state SOFIE-2504

Julusian:

feature

TSR is using v2.4 of atem-connection, this works perfectly for us, but is a little out of date. This means that when implementing new things, that ends up being done once for v3 and then backported (which sometimes results in almost completely rewriting the change from scratch)

This updates TSR to be using v3 of atem-connection. This includes the ability to control various additional areas of the atem (not yet exposed in TSR), and reworks some areas to be simpler or translate better to models released since v2.0 of atem-connection.

* **Other information**:

atem-connection is also a nrkno project, but a large amount of v3 was contributed by the community (some of which is also part of the nrkno team, but not done on nrkno time)


packages/timeline-state-resolver/src/integrations/atem/diffState.ts
```diff
@@ -0,0 +1,106 @@
...
+import {
+	Mapping,
+	SomeMappingAtem,
+	MappingAtemAuxilliary,
+	MappingAtemType,
+	MappingAtemAudioChannel,
+	Mappings,
+} from 'timeline-state-resolver-types'
+
+export function createDiffOptions(mappings: Mappings): DeepComplete<Diff.SectionsToDiff> {
```
nytamin, line 12:
```suggestion
/**
  Returns an option object to be passed into AtemState.diffStates().
  Based on the mappings, these options enables/disables certain areas-of-interest in the diff atem state.
*/
export function createDiffOptions(mappings: Mappings): DeepComplete<Diff.SectionsToDiff> {
```

packages/timeline-state-resolver/src/integrations/atem/index.ts
```diff
@@ -179,23 +176,9 @@ export class AtemDevice
...
-			)
-			.map((mapping) => mapping.options.index)
-
-		for (let i = 0; i < noOfAuxes; i++) {
-			if (!auxMappings.includes(i)) {
-				oldAtemState.video.auxilliaries[i] = undefined
-				newAtemState.video.auxilliaries[i] = undefined
-			}
-		}
+		const diffOptions = createDiffOptions(mappings)
```
nytamin, line 179: Question: Should we add a device option for enabling/disabling partial control?

packages/timeline-state-resolver/src/integrations/atem/stateBuilder.ts
```diff
@@ -123,8 +116,20 @@ export class AtemStateBuilder {
...
 		if (typeof mapping.index !== 'number' || mapping.index < 0) return
 
-		const stateMixEffect = AtemStateUtil.getMixEffect(this.#deviceState, mapping.index)
-		deepExtend(stateMixEffect, _.omit(content.me, 'upstreamKeyers'))
+		const stateMixEffect = deepMerge(
+			AtemStateUtil.getMixEffect(this.#deviceState, mapping.index),
+			_.omit(content.me, 'upstreamKeyers', 'transitionPosition')
+		)
+		this.#deviceState.video.mixEffects[mapping.index] = stateMixEffect
+		if (typeof content.me.transitionPosition === 'number') {
```
nytamin, line 124: Minor readability note:
To me, `typeof content.me.transitionPosition === 'number'` implies that there are other possible types, a `content.me.transitionPosition !== undefined` is slightly less suprising in this case, I think.

nytamin (approved):

packages/timeline-state-resolver/src/integrations/atem/index.ts
Julusian: This is currently intended to behave identically to previous versions of TSR in what it will control.
Long term, we should make this more configurable, but I don't want to do that as part of this PR.

PR #290, opened 2023-09-14
Fix: refactor TCP send device to the new device structure (SOFIE-2495)

nytamin:
This PR update the TCPSend device to the new device structure.

(Note: this PR is based upon https://github.com/nrkno/sofie-timeline-state-resolver/pull/287)


packages/timeline-state-resolver/src/integrations/tcpSend/tcpConnection.ts
```diff
@@ -0,0 +1,154 @@
...
+					this._setConnected(true)
+				})
+				this._tcpClient.on('close', () => {
+					this._setConnected(false)
+				})
+				this._tcpClient.on('end', () => {
+					this._setConnected(false)
+				})
+			}
+			const tcpClient: Socket = this._tcpClient
```
Julusian, line 51: I would move this chunk outside of the `new Promise()`, to minimise the amount of code inside the promise constructor. I view the promise constructor as a dangerous region of code, as it is easily possible to create bugs that will cause it to never resolve.

The `if (!this.connected) {` below could also be moved outside of the promise constructor

packages/timeline-state-resolver/src/integrations/tcpSend/tcpConnection.ts
```diff
@@ -0,0 +1,154 @@
...
+
+			if (!this.connected) {
+				tcpClient.connect(activeOptions.port, activeOptions.host, () => {
+					resolve(tcpClient)
+				})
+				setTimeout(() => {
```
Julusian, line 57: should there also be a listener for `error` here, to reject if the connection fails with a better failure than the timeout?

packages/timeline-state-resolver/src/integrations/tcpSend/tcpConnection.ts
```diff
@@ -0,0 +1,154 @@
...
+		tcpClient.write(Buffer.from(message, this.activeOptions.bufferEncoding))
+	}
+
+	private _setConnected(connected: boolean) {
+		if (this._connected !== connected) {
+			this._connected = connected
+			this._connectionChanged()
+
+			if (!connected) {
+				this._tcpClient = null
```
Julusian, line 122: It would be a good idea to do some cleanup of `_tcpClient` before forgetting it, to ensure it is closed and can be garbage collected.

packages/timeline-state-resolver/src/integrations/tcpSend/index.ts
```diff
@@ -1,282 +1,89 @@
...
-		await this._disconnectTCPClient()
-		await this._connectTCPClient()
-
-		return {
-			result: ActionExecutionResultCode.Ok,
-		}
-	}
+import { t } from '../../lib'
+import _ = require('underscore')
+import { error } from 'console'
```
Julusian, line 15: This doesnt look intentional

packages/timeline-state-resolver/src/integrations/tcpSend/index.ts
```diff
@@ -285,131 +92,98 @@ export class TCPSendDevice extends DeviceWithState<TSCSendState, DeviceOptionsTC
...
+					commandName: 'manual',
+					content: cmd,
+					layer: '',
+				},
+			})
+		} catch (e) {
+			this.context.emitWarning('Manual TCP command failed: ' + JSON.stringify(cmd))
+			return {
+				result: ActionExecutionResultCode.Error,
+				response: t('Error when sending TCP command: {{errorMessage}}', { errorMessage: `${error}` }),
```
Julusian, line 176:
```suggestion
				response: t('Error when sending TCP command: {{errorMessage}}', { errorMessage: `${e}` }),
```

This is not using the correct variable

Julusian (approved):

PR #298, opened 2023-10-13
Action return types

nytamin:
(This is a PR from SuperFly.tv)

## Background

We have an application, [SuperConductor](https://github.com/SuperFlyTV/SuperConductor), which uses TSR as its backend playout engine.
SuperConductor needs to talk to the devices to retrieve certain data, however currently it has to open a secondary connection to the devices, sine TSR doesn't expose any direct methods that it can use.

With this PR I'd like to propose a way to solve this issue.

### Example of use case

SuperConductor needs the list of media in CasparCG media folder. To retrieve this, it runs `casparcg.cls()`.

## Summary of changes

* Adds `returnData` to the actions schema.
* Modify `schema-types.mjs` script to generate types for both the `payload` and `returnData` along with helper types to use these in the devices.
* Adds `listMedia` action to CasparCG device as an example implementation

## Other information

TODO:

- [x] Draft schema changes
- [x] Example implementation (`CasaparCG.listMedia`)
- [x] Open PR and invite discussion
- [x] Schema approved after discussion
- [x] Test and verify example implementation in QuickTSR
- [ ] Merge PR
- [ ] Open secondary PR with more actions needed in SuperConductor

### DIFF
--- a/packages/timeline-state-resolver/scripts/schema-types.mjs
+++ b/packages/timeline-state-resolver/scripts/schema-types.mjs
@@ -95,8 +95,6 @@ for (const dir of dirs) {
 
 	const dirId = capitalise(dir)
 
-	console.log(dirId)
-
 	let output = ''
 
 	const generatedSchemaDirectory = path.join(basePathOfDereferencedShemas, dir)
@@ -194,7 +192,7 @@ for (const dir of dirs) {
 				const actionDefinition = {
 					id: action.id,
 					payloadId: undefined,
-					returnDataId: undefined
+					resultId: undefined
 				}
 				actionDefinitions.push(actionDefinition)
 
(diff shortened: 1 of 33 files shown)


packages/timeline-state-resolver/src/$schemas/action-schema.json
```diff
@@ -31,25 +34,54 @@
...
+							"type",
+							"properties"
+						],
+						"allOf": [
+							{
+								"$ref": "https://json-schema.org/draft/2020-12/schema"
+							}
+						]
+					},
+					"returnData": {
```
mint-dewit, line 55: `result` would be a slightly more elegant name I think?

packages/timeline-state-resolver/scripts/schema-types.mjs
```diff
@@ -87,6 +93,10 @@ let baseMappingsTypes = []
 for (const dir of dirs) {
 	const dirPath = path.join(basePath, dir)
 
+	const dirId = capitalise(dir)
+
+	console.log(dirId)
```
mint-dewit, line 98: Stray log line?

packages/timeline-state-resolver/src/integrations/casparCG/index.ts
```diff
@@ -729,6 +734,34 @@ export class CasparCGDevice extends DeviceWithState<State, DeviceOptionsCasparCG
...
+		if (result.error)
+			return {
+				result: ActionExecutionResultCode.Error,
+				response: t(`Error message from CasparCG: {{message}}`, { message: `${result.error}` }),
+			}
+
+		const request = await result.request
+
+		if (request.responseCode === 200) {
+			// TODO: implement return data
```
mint-dewit, line 753: todo here :)

mint-dewit (commented):
This is a perfect fit for the actions model.

The need to do so many type assertions in the executeAction implementation is quite clunky though, but it looks like that will go away as more integrations will be moved over to the new state handler form.

mint-dewit (approved):

PR #299, opened 2023-10-16
feat: VizMSE action to send clear-commands (configured on the device …

jesperstarkar:
VizMSE action to send clear-commands (configured on the device ettings) to all Engines in the Profile

Feature

You can only achieve `clearEngines` on `makeReady`.

Exposes a TSR action for `clearAllEngines` which clears all Engines known in the Profile.

* **Other information**:

### DIFF
--- a/packages/timeline-state-resolver/src/integrations/vizMSE/$schemas/actions.json
+++ b/packages/timeline-state-resolver/src/integrations/vizMSE/$schemas/actions.json
@@ -7,6 +7,13 @@
 			"description": "Purges all elements from the viz-rundown",
 			"destructive": true,
 			"timeout": 10000
+		},
+		{
+			"id": "clearAllEngines",
+			"name": "Clear Engines",
+			"description": "Clears all Engines in the Profile by sending the configured list of clear-commands",
+			"destructive": true,
+			"timeout": 10000
 		}
 	]
 }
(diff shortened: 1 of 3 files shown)


packages/timeline-state-resolver/src/integrations/vizMSE/$schemas/actions.json
```diff
@@ -7,6 +7,13 @@
 			"description": "Purges all elements from the viz-rundown",
 			"destructive": true,
 			"timeout": 10000
+		},
+		{
+			"id": "clearAllEngines",
+			"name": "Clear Engines",
+			"description": "Clears all Engines in the Profile by sending the configured list of clear-commands",
+			"destructive": true,
```
jesperstarkar, line 15: I'm not sure what `destructive` means, so I yoloed on `true`

jesperstarkar: It is a UX thing. True is good.

nytamin: yes, `true` is correct here. `destructive` means that the action can affect playout (ie "use with caution").

packages/timeline-state-resolver/src/integrations/vizMSE/index.ts
```diff
@@ -214,12 +214,24 @@ export class VizMSEDevice extends DeviceWithState<VizMSEState, DeviceOptionsVizM
 	public async purgeRundown(clearAll: boolean): Promise<void> {
 		await this._vizmseManager?.purgeRundown(clearAll)
 	}
+	public async clearEngines(): Promise<void> {
+		await this._vizmseManager?.clearEngines({
+			type: VizMSECommandType.CLEAR_ALL_ENGINES,
+			time: this.getCurrentTime(),
```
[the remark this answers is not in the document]
nytamin: Should be fine, I don't think they are used atm.

nytamin (approved):
Looks good!

PR #307, opened 2023-11-30
fix: reset atem upon connection

mint-dewit:
This is a bug fix.

**Expected behaviour**

The atem integration will initiate correctly and enforce its state.

**Observed behaviour**

The atem integration does a state diff before it has connected to the actual device. This returns no commands. The State Handler than transitions to this state and expects it to be current.

Once the device is connected, a `resetResolver` is called. This does not reset the state itself causing it to diff against an incorrect current state and will again return no commands

**Proposed fix**

Add a method to update the current state in the State Handler on the context. In addition, chane the `resetState` context method to change the current state to `undefined`

### DIFF
--- a/packages/timeline-state-resolver/src/service/DeviceInstance.ts
+++ b/packages/timeline-state-resolver/src/service/DeviceInstance.ts
@@ -194,7 +194,7 @@ export class DeviceInstanceWrapper extends EventEmitter<DeviceInstanceEvents> {
 		this._logDebugStates = value
 	}
 
-	private _getDeviceContextAPI(): DeviceContextAPI {
+	private _getDeviceContextAPI(): DeviceContextAPI<any> {
 		return {
 			logger: {
 				error: (context: string, err: Error) => {
--- a/packages/timeline-state-resolver/src/service/DeviceInstance.ts
+++ b/packages/timeline-state-resolver/src/service/DeviceInstance.ts
@@ -242,6 +242,13 @@ export class DeviceInstanceWrapper extends EventEmitter<DeviceInstanceEvents> {
 			},
 
 			resetState: async () => {
+				await this._stateHandler.setCurrentState(undefined)
+				await this._stateHandler.clearFutureStates()
+				this.emit('resetResolver')
+			},
+
+			resetToState: async (state: any) => {
+				await this._stateHandler.setCurrentState(state)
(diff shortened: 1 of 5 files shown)


packages/timeline-state-resolver/src/service/DeviceInstance.ts
```diff
@@ -242,6 +242,13 @@ export class DeviceInstanceWrapper extends EventEmitter<DeviceInstanceEvents> {
...
 			resetState: async () => {
+				await this._stateHandler.setCurrentState(undefined)
+				await this._stateHandler.clearFutureStates()
+				this.emit('resetResolver')
+			},
+
+			resetToState: async (state: any) => {
+				await this._stateHandler.setCurrentState(state)
 				await this._stateHandler.clearFutureStates()
 				this.emit('resetResolver')
```
Julusian, line 253: This isn't something to fix now, but doesn't calling `this.emit('resetResolver')` result in clearing the state of every device and rebuilding it all from the timeline again?
That feels like a rather heavy operation to do whenever a device manages to connect, especially as the states returned by the resolver will not be different to before.
Perhaps something in each service could cache the states as given by the resolver, so that `resetState` and `resetToState` can reuse those instead?

packages/timeline-state-resolver/src/service/device.ts
```diff
@@ -148,6 +152,9 @@ export interface DeviceContextAPI {
 
 	timeTrace: (trace: FinishedTrace) => void
 
-	/** Reset the state of the State */
+	/** Reset the tracked device state to undefined and notify the conductor to reset the resolver */
 	resetState: () => Promise<void>
+
+	/** Reset the tracked device state to "state" and notify the conductor to reset the resolver */
+	resetToState: (state: any) => Promise<void> // todo - types?
```
Julusian, line 159: Unresolved TODO
I agree it would be good to have the type of the state to be known here. It will probably be a bit sprawling, but should be easy enough to do as a generic parameter to this interface

Julusian (approved):

packages/timeline-state-resolver/src/service/device.ts
mint-dewit: My initial worry was that we already use a lot of generics and it does become a bit of a sprawl but in this case it's only a few places that need to be aware actually.

PR #308, opened 2023-12-04
feat: atem command batching SOFIE-2549

Julusian:
Blocked by https://github.com/nrkno/sofie-atem-connection/pull/157

feature

Commands are sent to an ATEM one at a time, one per packet.

For each state in time, the commands will be sent to the atem as a single batch (or packed into as few packets as possible).

This will help with responsiveness, as there is a limited number of packets allowed to be in flight at once, and it is not uncommon for 10+ commands to have the same start time.  

* **Other information**:


nytamin (approved):
Looks good from an initial read through.

PR #313, opened 2023-12-13
chore: enable strictFunctionTypes and `strictPropertyInitialization` typescript rules

Julusian:
## About the Contributor

## Type of Contribution

This is a: Code improvement

## New Behavior

This enables and fixes up some errors resulting from the additional checking rules.
These rules were disabled a while ago `to allow for gradual migration project`

## Status

### DIFF
--- a/packages/timeline-state-resolver/src/devices/deviceContainer.ts
+++ b/packages/timeline-state-resolver/src/devices/deviceContainer.ts
@@ -11,11 +11,16 @@ export { DeviceContainerEvents }
  * names and id's) to prevent a costly round trip over IPC.
  */
 export class DeviceContainer<TOptions extends DeviceOptionsBase<any>> extends BaseRemoteDeviceIntegration<TOptions> {
-	protected _device: ThreadedClass<Device<TOptions>>
-	public onChildClose: () => void | undefined
+	protected readonly _device: ThreadedClass<Device<TOptions>>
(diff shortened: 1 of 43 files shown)


packages/timeline-state-resolver/src/devices/deviceContainer.ts
```diff
@@ -11,8 +11,8 @@ export { DeviceContainerEvents }
  * names and id's) to prevent a costly round trip over IPC.
  */
 export class DeviceContainer<TOptions extends DeviceOptionsBase<any>> extends BaseRemoteDeviceIntegration<TOptions> {
-	protected _device: ThreadedClass<Device<TOptions>>
-	public onChildClose: () => void | undefined
+	protected _device!: ThreadedClass<Device<TOptions>>
```
nytamin, line 14: Perhaps we should instead add this to as an argument to the construtor, and in `static create()`:

```typescript
const device = await threadedClass<Device<TOptions>, TCtor>(
	orgModule,
	orgClassExport,
	[deviceId, deviceOptions, getCurrentTime] as any, // TODO types
	threadConfig
)
const container = new DeviceContainer(device, deviceOptions, threadConfig)
```

nytamin (approved):
It would be slightly safer and easier to read if we added a comment on all the `!:`'s, like so:
```typescript
/** assigned in init() */
private options!: HTTPSendOptions
```

PR #315, opened 2023-12-18
refactor: A series of refactorings and fixes in vMix integration

ianshade:
## About the Contributor

## Type of Contribution

This is a: 

Code improvement that also tackles some issues that can be considered as bugs

## Current Behavior

- Total control of the features available in this integration is assumed, causing TSR to reset properties that may be intended to be controlled manually or left at the defaults stored in the Preset. This causes several conflicts:
  - audio properties of inputs are reset even when not intended to be controlled by the TSR, but for example with other tools like Sisyfos (even though separate mappings for audio already exist, the state is combined together with other input properties)
  - layer (overlay / multi view) properties of inputs are incorrectly parsed from vMix and reset at unexpected times
  - output assignments are reset
  - streaming, recording, ..., are turned off
- Because of the retry functionality, commands resetting the Alpha value are sent to all inputs every time the real state is polled (the value is not available in the XML returned by vMix)
- Inputs previously added by TSR are incorrectly treated as _fixed_ inputs when TSR is restarted
- Usage of deepMerge causes duplication in some arrays
- Some debug events are emitted without respecting the `debugLogging` flag

## New Behavior

- The code is split into multiple classes
- Similarly to the TriCaster integration, a resource is controlled (and may receive TSR-default state) on an opt-in basis, only when a layer mapping associated with it exists. This includes:
  - inputs (each one individually)
  - audio channels (each one individually)
  - outputs (each one individually)
  - overlays (each one individually)
  - recording
  - streaming
  - external
  - mixes (each one individually)
- Readme file describing implementation details and limitations is added
- The retry functionality is refactored. Only relevant properties of inputs are synced, not deeply merging the whole state anymore. An edge case potentially causing duplicate commands to be sent due to a response to polling arriving after commands from a new resolving are sent is (hopefully) solved.
- Commands setting Alpha are not re-sent repeatedly as a result of polling
- Input layers (overlay / multi view) parsing is fixed
- Inputs added by TSR are handled separately from the pre-existing inputs for an improved recovery when restarting the TSR. Note: this might need further improvements that are out of the scope for this PR and partially depend on features already existing in release51.
- Debug logging is done through `emitDebug`. Note: this needs further improvements to separate commands from other logs

## Other Information

Many of the changes are based on https://github.com/nrkno/sofie-timeline-state-resolver/pull/235.
Some TODOs and one uncommented test case are left for when this will be merged with release51.

## Status


mint-dewit (approved):
I'm happy to accept this into R50 since they are constrained to just the VMix integration which is not in use by NRK right now. 

Before merging I just want to see the naming of the files be camelCase for consistency.

PR #316, opened 2024-01-03
OBS refactoring (Sofie 2486)

mint-dewit:
- Upgrades to obs-websocket-js v5 in order to gain compatibility with OBS 29+
- Adds some extra options for controlling transforms and media

Currently using a github repo instead of the NPM package awaiting this PR https://github.com/obs-websocket-community-projects/obs-websocket-js/pull/347


packages/timeline-state-resolver/src/integrations/obs/diff.ts
```diff
@@ -0,0 +1,359 @@
+import _ = require('underscore')
+import { OBSCommandWithContext, OBSCommandWithContextTyped } from '.'
+import { OBSDeviceState, OBSScene, OBSSceneItem, OBSInputState } from './state'
+import { OBSRequest as OBSRequestName } from 'timeline-state-resolver-types'
```
nytamin, line 4: This rename is unnecessary (?)

packages/timeline-state-resolver-types/src/integrations/obs.ts
```diff
@@ -1,35 +1,43 @@
 import { DeviceType } from '..'
 
 export enum OBSRequest {
```
nytamin, line 3: This enum is only ever used in TSR, not externally. So we could move it into the TSR package instead.

(
While we're at it, add this comment as well:
`/** values are derived from keyof OBSRequestTypes in obs-websocket-js */`
)

nytamin (approved):

PR #320, opened 2024-01-29
fix(vMix): fragmented message handling SOFIE-2932

ianshade:
## About the Contributor

## Type of Contribution

This is a: 

Bug fix / Code improvement

## Current Behavior

It could happen that incomplete XML message would be attempted to be processed by the 'data' event handler, causing an error to be thrown. As a result the old, incomplete data got indefinitely stuck in the `_unprocessedLines` of the `BaseConnection`.

## New Behavior

The logic for processing incoming data is extracted, refactored and covered by unit tests.
Handling of messages that may be fragmented over multiple TCP packets is fixed.
Errors thrown during the message handler are no longer interrupting the processing.
When reconnecting, previous and possibly incomplete data is discarded.

## Other Information

An assumption about the XML message being always terminated by a newline was made. This is not explicitly specified in the documentation, but the documentation says the following:
> All messages are UTF8 encoded and \r\n terminated unless specified otherwise.

> Optional binary data is the exact length specified in <status/length> including additional terminating characters such as \r\n
which are always appended to the end of multiline text such as from XMLTEXT.

(source: https://www.vmix.com/help25/index.htm?TCPAPI.html)

## Status

### DIFF
--- a/packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
+++ b/packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
@@ -29,77 +29,76 @@ export class VMixResponseStreamReader extends EventEmitter<ResponseStreamReaderE
 	processIncomingData(data: Buffer) {
 		const string = this._lineRemainder + data.toString('utf-8')
 		this._lineRemainder = ''
-		const newLines = string.split('\r\n')
+		const lines = string.split('\r\n')
 
-		if (newLines[newLines.length - 1] === '') {
-			// we're lucky and the data ends with a newline
-			this._unprocessedLines.push(...newLines)
+		if (lines[lines.length - 1] === '') {
+			// the data ends with a newline
+			this._unprocessedLines.push(...lines)
 		} else {
-			const incompleteLine = newLines.pop()
-			this._unprocessedLines.push(...newLines)
+			const incompleteLine = lines.pop()
+			this._unprocessedLines.push(...lines)
 			// we need to keep the remaining incomplete line
 			if (incompleteLine != null) {
-				this._lineRemainder += incompleteLine
+				this._lineRemainder = incompleteLine
 			}
 		}
 
-		lineProcessing: while (this._unprocessedLines.length > 0) {
-			const firstLine = this._unprocessedLines[0]
-			const result = RESPONSE_REGEX.exec(firstLine)
-			let processedLines = 0
+		let lineToProcess: string | undefined
+
+		while ((lineToProcess = this._unprocessedLines.shift()) != null) {
+			const result = RESPONSE_REGEX.exec(lineToProcess)
 
 			if (result && result.groups?.['response']) {
+				const responseLen = parseInt(result?.groups?.['response'])
+
 				// create a response object
-				// Add 2 to account for the space between `command` and `response` as well as the newline after `response`.
-				const responseHeaderLength = result.groups?.['command'].length + result.groups?.['response'].length + 2
-				if (Number.isNaN(responseHeaderLength)) {
-					break lineProcessing
-				}
-				const responseLen = parseInt(result?.groups?.['response']) - responseHeaderLength
 				const response: Response = {
 					command: result.groups?.['command'],
 					response: (Number.isNaN(responseLen) ? result.groups?.['response'] : 'OK') as Response['response'],
 					message: result.groups?.['responseMsg'],
 					body: undefined as undefined | string,
 				}
-				processedLines++
 
 				// parse payload data if there is any
 				if (!Number.isNaN(responseLen)) {
-					let len = responseLen
-					const lines: string[] = []
-
-					while (len > 0) {
-						const l = this._unprocessedLines[lines.length + 1] // offset of 1 because first line is not data
-						if (l === undefined) {
-							// we have not received all the data from server, break line processing and wait for more data
-							break lineProcessing
-						}
-
-						len -= l.length + 2
-						lines.push(l)
+					const payloadData = this.processPayloadData(responseLen)
+					if (payloadData == null) {
+						this._unprocessedLines.unshift(lineToProcess)
+						break
+					} else {
+						response.body = payloadData
 					}
-					response.body = lines.join('')
-					processedLines += lines.length
 				}
 
 				// now do something with response
 				try {
 					this.emit('response', response)
 				} catch (e) {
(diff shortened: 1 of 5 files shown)


Julusian:
Hello!

Thank you for contributing to the Sofie Project!

If you haven’t already, please give our [contribution guidelines](https://nrkno.github.io/sofie-core/docs/for-developers/contribution-guidelines/) a read. We will reach out to you if we have any questions on this.

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
```diff
@@ -0,0 +1,104 @@
...
+		const lines = string.split('\r\n')
+
+		if (lines[lines.length - 1] === '') {
+			// the data ends with a newline
+			this._unprocessedLines.push(...lines)
+		} else {
+			const incompleteLine = lines.pop()
+			this._unprocessedLines.push(...lines)
+			// we need to keep the remaining incomplete line
+			if (incompleteLine != null) {
```
Julusian, line 41:
```suggestion
			if (incompleteLine) {
```
`!= null` is an odd choice here, as im pretty sure `pop()` returns `string | undefined`. May as well be a falsey check though

Also I think sonarcloud will complain about usage of `!=` and `==`. So they are better to avoid without a good reason for them

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
```diff
@@ -0,0 +1,104 @@
...
+		this._unprocessedLines = []
+		this._lineRemainder = ''
+	}
+
+	processIncomingData(data: Buffer) {
+		const string = this._lineRemainder + data.toString('utf-8')
+		this._lineRemainder = ''
+		const lines = string.split('\r\n')
+
+		if (lines[lines.length - 1] === '') {
```
Julusian, line 34: I feel like this could be written simpler by doing a `const lastNewlineIndex = string.lastIndexOf('\r\n')`.
then it would be something like:
```
const lastNewlineIndex = string.lastIndexOf('\r\n')
if (lastNewLineIndex === -1) {
  this._lineRemainder = string
  return
} else {
  const lines = string.slice(0, lastNewLineIndex).split('\r\n')
  this._unprocessedLines.push(...lines)
  this._lineRemainder = string.slice(lastNewLineIndex + 2)
}

```
To me this feels more natural, but maybe it isnt any simpler actually.

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
```diff
@@ -0,0 +1,104 @@
...
+			this._unprocessedLines.push(...lines)
+			// we need to keep the remaining incomplete line
+			if (incompleteLine != null) {
+				this._lineRemainder = incompleteLine
+			}
+		}
+
+		let lineToProcess: string | undefined
+
+		while ((lineToProcess = this._unprocessedLines.shift()) != null) {
```
Julusian, line 48:
```suggestion
		while ((lineToProcess = this._unprocessedLines.shift()) !== undefined) {
```

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
```diff
@@ -0,0 +1,104 @@
...
+					if (payloadData == null) {
+						this._unprocessedLines.unshift(lineToProcess)
+						break
+					} else {
+						response.body = payloadData
+					}
+				}
+
+				// now do something with response
+				try {
```
Julusian, line 74: I feel like it would be better to put this `try` above the `const response: Response` (or perhaps just inside of the `while`?) so that if there are other bugs here it wont break the loop, and will end up ignoring a message instead of getting stuck again

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
```diff
@@ -0,0 +1,104 @@
...
+			}
+		}
+	}
+
+	private processPayloadData(responseLen: number): string | null {
+		const processedLines: string[] = []
+
+		while (responseLen > 0) {
+			const line = this._unprocessedLines[processedLines.length]
+			if (line == null) {
```
Julusian, line 93:
```suggestion
			if (line === undefined) {
```

Julusian (approved):

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
ianshade: I'm personally a huge fan of `== null` and `!= null` as a way of checking whether a value is/isn't nullish. If eslint has a rule that allows it, wouldn't sonarcloud too? https://eslint.org/docs/latest/rules/eqeqeq#smart

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
ianshade: I think that would be going too far in not trusting our own code.

Julusian: It is a bit, but I still like doing it. I don't think it makes the code any less readable and helps reinforce the approach of 'if something errors ignore it and continue'.
Im not opposed to leaving it as is though as the real world benefit is likely non existent

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
Julusian: I didn't realise that these checks had smart and useful behaviour. I assume our eslint must be configured to allow these ones otherwise it would have complained, so maybe sonar will too. 
For now lets leave them as is, I've added it to our internal devmeet agenda, in case anyone feels strongly against this.

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
Julusian: I like it. nice and short and easy to follow

packages/timeline-state-resolver/src/integrations/vmix/vMixResponseStreamReader.ts
ianshade: Updated as requested

PR #328, opened 2024-05-03
Refactor connection management

mint-dewit:
## Type of Contribution

This is a  Code improvement

## Current Behavior
Library consumers must keep track of how devices are created and initialised and are responsible for managing this. In addition, a device may not finish initialising until the first connection has been made.

## Proposed Behavior
A Connection is an instance of an Integration Implementation (previously known as devices, this terminology proved to be a bit confusing as "TSR devices" connected to actual devices). TSR shall be responsible for creating and initialising the connections based upon a config provided by the library user. An integration must ensure to not get stuck during the initialisation.

## Discussion
~~I think there may actually some use for devices not being initialised immediately as it prevents the timeline states from being sent into the integration when it isn't ready yet. At the same time, integrations should handle this "disconnected" state gracefully. Perhaps some additional life cycle hooks are required (i.e. to rediff the state from empty)~~

## Status
This PR is part investigation and part implementation. Feedback is welcome at this time.

### DIFF
--- a/packages/timeline-state-resolver/src/conductor.ts
+++ b/packages/timeline-state-resolver/src/conductor.ts
@@ -1,7 +1,7 @@
 import * as _ from 'underscore'
 import { getResolvedState, ResolvedTimeline, ResolvedTimelineObjectInstance, TimelineObject } from 'superfly-timeline'
 import { EventEmitter } from 'eventemitter3'
-import { MemUsageReport, threadedClass, ThreadedClass, ThreadedClassConfig, ThreadedClassManager } from 'threadedclass'
+import { MemUsageReport, threadedClass, ThreadedClass, ThreadedClassManager } from 'threadedclass'
 import PQueue from 'p-queue'
 import * as PAll from 'p-all'
 import PTimeout from 'p-timeout'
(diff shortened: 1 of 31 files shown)


packages/timeline-state-resolver/src/service/ConnectionManager.ts
```diff
@@ -0,0 +1,395 @@
...
+		// set up event handlers
+		this._setupDeviceListeners(id, container)
+
+		this._connections.set(id, container)
+		this.emit('connectionAdded', id, container)
+
+		// trigger device init
+		this._handleConnectionInitialisation(id, container).catch(() => {
+			this.emit('error', 'Device ' + id + ' failed to initialise')
+			this._connections.delete(id) // todo - this will cause device to be recreated next, which may cause a spiral?
```
Julusian, line 171: Is more cleanup needed?
This is discarding the thread, but the thread is probably still running at this point?

packages/timeline-state-resolver/src/conductor.ts
```diff
@@ -172,7 +167,7 @@ export class Conductor extends EventEmitter<ConductorEvents> {
 
 	private _options: ConductorOptions
 
-	private devices = new Map<string, BaseRemoteDeviceIntegration<DeviceOptionsBase<any>>>()
+	public readonly connectionManager = new ConnectionManager()
```
Julusian, line 170: should this be public, or should select methods be re-exposed on this conductor?

Julusian (commented):
There are still a bunch of todos that need resolving, but I like the approach taken here and don't have any major problems with it.

I haven't read changes made to devices, but I suspect that some care will be needed in them. I know that some of them are doing assertions on when properties are defined, as they get defined by init, but now that init resolves immediately, will those assumptions still be correct?

mint-dewit:
This is pretty much ready and I'm leaving this in state I'm happy with but the unit tests appear to be very inconsistent for reasons that elude me so that will require some attention

nytamin (approved):

PR #342, opened 2024-08-16
feat: http url interpolation SOFIE-3310

Julusian:
## About the Contributor

## Type of Contribution

This is a Feature

## New Behavior
This allows for defining the urls for the httpsend device, as well as urls passed to the renderers in sofieChef and casparcg, in a format that allows for taking fragments from the datastore.

For example an object could look something like:
```
{
  deviceType: 'httpSend',
  type: 'post',
  url: {
    key: `http://example.com/{{subPath}?apiKey={{apiKey}}`,
    args: {
      subPath: 'my-default', // Overriden by the datastore
      apiKey: '', // Taken from datastore
    }
  },
  $references: {
    'url.args.subPath': {
      datastoreKey: 'httpSubpath',
      overwrite: true,
    },
    'url.args.apiKey': {
      datastoreKey: 'httpApiKey',
      overwrite: false,
    }
  }
}
```
In this example, the 'apiKey' will be taken from the datastore, avoiding being baked into the rundown.
And the 'subPath' being taken from the datastore, means it can be changed while playing the rundown more easily.

The syntax of the interpolation matches the translations format we use elsewhere in Sofie.

This has been added to the httpSend and sofieChef devices, and to casparcg for the `html` content type

## Status

### DIFF
--- a/packages/timeline-state-resolver-types/src/index.ts
+++ b/packages/timeline-state-resolver-types/src/index.ts
@@ -172,3 +172,10 @@ export enum ActionExecutionResultCode {
 	Error = 'ERROR',
 	Ok = 'OK',
 }
+
+export interface StringInterpolation {
+	key: string
+	args?: {
+		[k: string]: any
+	}
+}
--- a/packages/timeline-state-resolver-types/src/index.ts
+++ b/packages/timeline-state-resolver-types/src/index.ts
@@ -173,8 +173,11 @@ export enum ActionExecutionResultCode {
 	Ok = 'OK',
 }
 
-export interface StringInterpolation {
+/** This resolves to a string, where parts can be defined by the datastore */
+export interface TemplateString{
+	/** The string template. Example: "http://google.com?q={{searchString}}" */
 	key: string
+	/** Values for the arguments in the key string. Example: { searchString: "TSR" } */
 	args?: {
 		[k: string]: any
 	}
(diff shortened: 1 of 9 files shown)


packages/timeline-state-resolver-types/src/index.ts
```diff
@@ -172,3 +172,10 @@ export enum ActionExecutionResultCode {
 	Error = 'ERROR',
 	Ok = 'OK',
 }
+
+export interface StringInterpolation {
+	key: string
+	args?: {
+		[k: string]: any
+	}
+}
```
nytamin, line 181: Slightly confused by the naming, perhaps rename to `TemplateString`? This could also add some documentation?

```suggestion
/** This resolves to a string, where parts can be defined by the datastore */
export interface TemplateString{
	/** The string template. Example: "http://google.com?q={{searchString}}" */
	key: string
	/** Values for the arguments in the key string. Example: { searchString: "TSR" } */
	args?: {
		[k: string]: any
	}
}
```

packages/timeline-state-resolver/src/lib.ts
```diff
@@ -307,3 +308,25 @@ export function actionNotFoundMessage(id: never): ActionExecutionResult<any> {
 export function cloneDeep<T>(input: T): T {
 	return klona(input)
 }
+
+/**
+ * Interpolate a translation style string
+ */
+export function interpolateTranslation(key: string, args: { [key: string]: any } | undefined): string {
```
nytamin, line 315: This function could use a unit test

nytamin (changes requested):

PR #346, opened 2024-09-05
fix(HTTPSend): Allow HTTPSend commands to be sent in separate sequential queues (SOFIE-3431)

jstarpl:
## About the Contributor

## Type of Contribution

This is a: 

Bug fix

## Current Behavior

All commands in an HTTPSend device are sent sequentially, ignoring the `queueId` property.

## New Behavior

HTTPSend device again supports the queueId property to allow separate sequential request queues.

## Testing Instructions

* Ensure that HTTP integration still works as expected
* See if multiple parallel HTTP request can be made when using the queueId

## Status

### DIFF
--- a/packages/timeline-state-resolver/src/integrations/quantel/connection.ts
+++ b/packages/timeline-state-resolver/src/integrations/quantel/connection.ts
@@ -591,11 +591,10 @@ export class QuantelManager extends EventEmitter {
 		})
 	}
 	public clearAllWaitWithPort(portId: string) {
-		if (!this._waitWithPorts[portId]) {
-			_.each(this._waitWithPorts[portId], (fcn) => {
-				fcn(true)
-			})
-		}
+		if (!this._waitWithPorts[portId]) return
+		_.each(this._waitWithPorts[portId], (fcn) => {
+			fcn(true)
+		})
 	}
 	/**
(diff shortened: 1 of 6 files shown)


packages/timeline-state-resolver/src/integrations/quantel/connection.ts
```diff
@@ -591,11 +591,10 @@ export class QuantelManager extends EventEmitter {
 		})
 	}
 	public clearAllWaitWithPort(portId: string) {
-		if (!this._waitWithPorts[portId]) {
```
Julusian, line 594: There is a bug task already for this SOFIE-3419, as this is not a recent mistake and has always been there

jstarpl: I did not see any use of `clearAllWaitWithPort` but it seemed clearly wrong, so I just changed it now.

Julusian: Oh right yeah, this quantel one isn't called anymore due to a mistake in the refactoring. so I guess this can be in whatever state you prefer as it is currently dead code. It is also refactored in r52 too, so expect a minor merge conflict there

PR #352, opened 2024-10-03
Quantel: reimplement clearAllWaitWithPort (SOFIE-3551)

nytamin:
## About the Contributor

## Type of Contribution

This is a: Bug/functionality fix

## Current Behavior

There currently is a situation where:

1. Clip A is playing
2. Clip A is stopped, however it has an outTransition: DELAY set
  so the stop command wont be sent until a while later
3. Clip B starts playing (on same port)

The Quantel device might either
* Wait with playing B until the stop-command for A has executed.
* Start playing B, then stop it (because of the delayed stop-command)

## New Behavior

This PR has multiple fixes.

First, the funcitonality of `CommandExecutor` has been rewritten to support a mix of SALVO and SEQUENTIAL commands.

Secondly, the Quantel device has now added a CLEARWAITING command (of type SALVO) that is exececuted before any play, stop, clear, ect, command, to ensure that a delayed command is cancelled, whatever it was going to do will be replaced by the new command.

## Testing Instructions

* This needs testing with Quantel.
* This needs testing with all devices.

## Status

- [ ] The functionality has been tested by the author.


packages/timeline-state-resolver/src/service/commandExecutor.ts
```diff
@@ -18,11 +22,44 @@ export class CommandExecutor<DeviceState, Command extends CommandWithContext> {
...
+
+		// The execution logic is as follows:
+		// * When the commands sent into executeCommands() is called a BATCH.
+		//
+		// * Salvos should wait for Salvos from previous BATCH to have finished before executing
+		// * Salvos within the same BATCH are executed in parallel.
+		//
+		// * Sequentials should wait for all Salvos and Sequentials from previous BATCH to have finished before executing
+		// * Sequentials should be executed after all Salvos from the same BATCH have been executed
+		// * Sequentials within the same BATCH are executed in order
```
jstarpl, line 45: I am somewhat worried about this algorithm mixing commands from different batches. After all, Salvos from Batch N+1 have been generated assuming that Salvos _and_ Sequentials from Batch N have already been sent. I feel like each batch should be done, before we start sending commands for the next batch? They are effectively applying state, and thus each batch depends on the previous one to have been applied?

jstarpl (commented):
I don't see any problems with the implementation, but I'm a bit worried about the algorithm implemented.

packages/timeline-state-resolver/src/integrations/quantel/diff.ts
```diff
@@ -104,6 +105,36 @@ export function diffStates(
...
+
+	// If we run any play-command, we will need to cancel any delayed/waiting out-transitions for that port:
+	const portIdsToCancelWaiting: Set<string> = new Set()
+	for (const cmd of allCommands) {
+		if (
+			(cmd.command.type === QuantelCommandType.PLAYCLIP ||
+				cmd.command.type === QuantelCommandType.PAUSECLIP ||
+				cmd.command.type === QuantelCommandType.CLEARCLIP ||
+				cmd.command.type === QuantelCommandType.RELEASEPORT) &&
+			!cmd.command.fromLookahead
```
mint-dewit, line 117: I think this should be looking at the `port.notOnAir` property instead like it does in https://github.com/nrkno/sofie-timeline-state-resolver/blob/ff443bb9260e05288c0c7c96950233ca79d8a679/packages/timeline-state-resolver/src/integrations/quantel/diff.ts#L204

However I can see that in the original release50 implementation this wasn't the case either (but it wasn't triggering for clear and release there either)

packages/timeline-state-resolver/src/service/commandExecutor.ts
nytamin: Perhaps a solution to this is to write some really good documentation about the intention of the type of commands?

For my use case, I would need it to work this way, because:

* Batch 1:
  * Sequntial: Quantel stop command (delayed, so will wait and not resolve until after an `await sleep(x)` )
* Batch 2:
  * Salvo: cancel any waiting Quantel operations
  * Sequntial: Quantel play command

So in this case, the cancel-salvo _must not_ wait for the previous batch to finish.

---

The only other possible solution to this would be to handle specific situations like this internally in the device - but that goes agains the idea of the rewrite of the TSR devices where we try to unify the integration..

nytamin:
After a discussion with Jan we figured that we should be able to achieve an equivalent functionality by only using a uniquely-named queueId for the CancelWaiting command.
This would mean that we won't have to rework the commandExecutor and thus make a smaller fix.
I will close this PR and follow up with another one

nytamin:
replaced by https://github.com/nrkno/sofie-timeline-state-resolver/pull/379

PR #353, opened 2024-10-04
fix(sisyfos): Use nullish check rather than or on channel reset

PeterC89:
## About the Contributor
This pull request is posted on behalf of the BBC.

## Type of Contribution

Bug fix

## Current Behavior
When setting a full Sisyfos channel state falsy values (E.G. a fader level of 0) will cause the default to be sent instead of the desired value.

## New Behavior
The correct value is sent

## Testing Instructions
* Trigger a set channel command (E.G. with `triggerValue` on an individual channel) with a fader level of 0 and verify that Sisyfos receives 0 as expected

## Status
- [X] PR is ready to be reviewed.
- [X] The functionality has been tested by the author.


nytamin (approved):
I'll cherry-pick this bug fix into release51 as well

PR #367, opened 2025-02-11
feat(sisyfos): add `disableDefaults` option to sisyfos mappings

ianshade:
## About the Contributor

This pull request is posted on behalf of TV 2 Norge

## Type of Contribution

This is a: 

Feature 

## Current Behavior

Sisyfos integration enforces certain default values for mapped channels, even if no timeline objects are present on the timeline, or if the timeline objects have certain values omited. It is, however, a requirement for us to leave the fader level, gain and input selector where the user set it through the GUI of Sisyfos, and never change nor reset them from the automation, but still have the automation control PGM and MUTE.

## New Behavior

A new, optional, and backwards compatible option called `disableDefaults` is introduced to Sisyfos layer mappings. In the UI it is titled "Do not apply default state for this layer". When set to `true` on a particular mapping, no defaults will be applied for the mapped channel(s). Only values present at a given moment on the timeline (on this layer) will be sent out in commands. In this case any need for default/fallback values (when no other timeline object is present on such layer) shall be defined explicitly in the "baseline" - for Sofie Blueprint developers it might mean adding an extra timeline object to the Studio/Rundown Baseline.

## Testing Instructions

1. Define a mapping with this option enabled.
2. Define an empty timeline object on this layer mapping
3. Un-default the channel: Move the fader, gain, and make other adjustments to the mapped Channel(s) through the GUI of Sisyfos.
4. Start TSR and verify that the adjustments were not reset
5. Start adding properties to the timeline object, and see the properties of the channel changing to the values set in it

## Status

### DIFF
--- a/packages/timeline-state-resolver-types/src/generated/sisyfos.ts
+++ b/packages/timeline-state-resolver-types/src/generated/sisyfos.ts
@@ -14,15 +14,18 @@ export interface SisyfosOptions {
 export interface MappingSisyfosChannel {
 	channel: number
 	setLabelToLayerName?: boolean
+	disableDefaults?: boolean
 	mappingType: MappingSisyfosType.Channel
 }
 
 export interface MappingSisyfosChannelByLabel {
 	label: string
+	disableDefaults?: boolean
 	mappingType: MappingSisyfosType.ChannelByLabel
 }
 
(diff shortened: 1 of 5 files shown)


nytamin (approved):
Very nice addition, thanks!

PR #369, opened 2025-02-26
Feat: WebSocket Client

olzzon:
## About the Contributor
This is made on behalf of BBC

## Type of Contribution
This is a feature and code comment improvements 

## New Behavior
This adds support for a general WS socket device

## Status

- [ ] The functionality has been tested by the author.


packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+  }
+export type WebSocketTcpClientDeviceState = Timeline.TimelineState<TSRTimelineContent>
+
+  export class WebSocketTcpClientDevice extends Device<
+	WebSocketTCPClientOptions,
+	WebSocketTcpClientDeviceState,
+	WebSocketTcpCommand
+> {
+	// Use ! as the connection will be initialized in init:
+	private connection!: WebSocketTcpConnection
```
mint-dewit, line 24: This is unsafe
```suggestion
	private connection: WebSocketTcpConnection | undefined = undefined
```

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+
+	constructor(context: DeviceContextAPI<any>) {
```
mint-dewit, line 26: No need to explicitly have a constructor if it does nothing

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+
+	readonly actions = {
+		[WebsocketTcpClientActions.Reconnect]: async (_id: string) => {
+			await this.connection.connect()
+			return { result: ActionExecutionResultCode.Ok }
+		},
+		[WebsocketTcpClientActions.ResetState]: async (_id: string) => {
+			return { result: ActionExecutionResultCode.Ok }
+		},
+		[WebsocketTcpClientActions.SendWebSocketMessage]: async (_id: string, payload?: Record<string, any>) => {
```
mint-dewit, line 44: The typings for the payload are wrong, as per the `actions.json` it should be
```suggestion
		[WebsocketTcpClientActions.SendWebSocketMessage]: async (_id: string, payload?: SendWebSocketMessagePayload) => {
```

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+			if (!payload?.message) {
+				return { result: ActionExecutionResultCode.Error, response: { key: 'Missing message in payload' } }
+			}
+			await this.sendCommand(payload.message)
```
mint-dewit, line 48: `payload.message` is a string, and sendCommand expects a `WebSocketTcpCommand`
```suggestion
			await this.sendCommand({
				command: {
					type: TimelineContentTypeWebSocketTcpClient.WEBSOCKET_MESSAGE,
					message: payload.message,
					isBase64Encoded: false // this isn't clarified on the payload?
				},
				context: 'Sent through action',
				timelineObjId: ''
			})
```

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+			return { result: ActionExecutionResultCode.Ok }
+		},
+		[WebsocketTcpClientActions.SendTcpMessage]: async (_id: string, payload?: Record<string, any>) => {
```
mint-dewit, line 51: Same for this function as above

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+		const commands: WebSocketTcpCommand[] = []
+		for (const [layerName, timelineObject] of Object.entries(newState.layers)) {
+			if (timelineObject.content.deviceType !== DeviceType.WEBSOCKET_TCP_CLIENT) continue
+
+
+			// We should send the command whenever the timeline object content has been ADDED or CHANGED
+			let changeType = 'N/A'
+			if (!oldState.layers[layerName]) {
+				changeType = 'added'
+			} else if (JSON.stringify(oldState.layers[layerName].content) !== JSON.stringify(timelineObject.content)) {
```
mint-dewit, line 98: `oldState` and `oldState.layers[layerName]` could be undefined
```suggestion
			} else if (JSON.stringify(oldState?.layers[layerName]?.content) !== JSON.stringify(timelineObject.content)) {
```

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+		// tl obj: { layer: 'abc', content: { type: TimelineContentTypeWebSocketTcpClient::TCP_MESSAGE, message: 'hello'} }
+		// can be converted into: { tcpMessages: { abc: 'hello } }
+		//
+		// This is optional and for convenience only (like to simplify the diffing logic in diffStates())
+
+		return state
+	}
+
+	// ** Calculate Diffs of state and create the commands
+	public diffStates(oldState: WebSocketTcpClientDeviceState, newState: WebSocketTcpClientDeviceState): WebSocketTcpCommand[] {
```
mint-dewit, line 83:
```suggestion
	public diffStates(oldState: WebSocketTcpClientDeviceState | undefined, newState: WebSocketTcpClientDeviceState): WebSocketTcpCommand[] {
```

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+
+
+		const commands: WebSocketTcpCommand[] = []
+		for (const [layerName, timelineObject] of Object.entries(newState.layers)) {
+			if (timelineObject.content.deviceType !== DeviceType.WEBSOCKET_TCP_CLIENT) continue
+
+
+			// We should send the command whenever the timeline object content has been ADDED or CHANGED
+			let changeType = 'N/A'
+			if (!oldState.layers[layerName]) {
```
mint-dewit, line 96:
```suggestion
			if (!oldState?.layers[layerName]) {
```

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
```diff
@@ -0,0 +1,154 @@
...
+				commands.push({
+					command: {
+						type: TimelineContentTypeWebSocketTcpClient.TCP_MESSAGE,
+						message: timelineObject.content.message,
+						isBase64Encoded: timelineObject.content.isBase64Encoded
+					},
+					context: `${changeType} on layer "${layerName}"`,
+					timelineObjId: timelineObject.id,
+				})
+			}
```
mint-dewit, line 128: This entire block seems to be the same as doing
```suggestion
				commands.push({
					command: {
						type: timelineObject.content.type ,
						message: timelineObject.content.message,
						isBase64Encoded: timelineObject.content.isBase64Encoded
					},
					context: `${changeType} on layer "${layerName}"`,
					timelineObjId: timelineObject.id,
				})

```

packages/timeline-state-resolver/src/service/device.ts
```diff
@@ -78,19 +91,28 @@ export abstract class Device<DeviceOptions, DeviceState, Command extends Command
 export interface BaseDeviceAPI<DeviceState, Command extends CommandWithContext> {
 	/**
 	 * This method takes in a Timeline State that describes a point
-	 * in time on the timeline and returns a decice state that
-	 * describes how the device should be according to the timeline state
-	 *
+	 * in time on the timeline and converts it into a "device state" that
+	 * describes how the device should be according to the timeline state.
+	 * This is optional, and intended to simplify diffing logic in
```
mint-dewit, line 96: It's not optional, optional would mean the developer can omit the method entirely. Whether the implementation transforms the state at all is up to the developer though

packages/timeline-state-resolver/src/integrations/websocketTcpClient/$schemas/options.json
```diff
@@ -0,0 +1,43 @@
...
+            "ui:title": "Optional - TCP Host",
+            "description": "Optional TCP server host to connect to, e.g. '127.0.0.1' - some devices may have a TCP connection too"
+          },
+          "port": {
+            "type": "integer",
+            "ui:title": "TCP Port",
+            "description": "TCP server port to connect to"
+          }
+        },
+        "required": ["host", "port"]
```
mint-dewit, line 38: These host and port should be optional?

packages/timeline-state-resolver/src/integrations/websocketTcpClient/$schemas/options.json
```diff
@@ -0,0 +1,43 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "WebSocket & TCP Client Options",
+    "type": "object",
+    "properties": {
+      "webSocket": {
```
mint-dewit, line 6: Missing a `"additionalProperties": false` for webSocket and for tcp objects

packages/timeline-state-resolver/src/integrations/websocketTcpClient/__tests__/websocketTcpClient.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+		MockWebSocketTcpConnection.prototype.sendTcpMessage.mockImplementation(() => {})
+
+		// Initialize device
+		await device.init( options )
+	})
+
+	afterEach(() => {
+		//Are there something like??:
+		//mockTime.dispose()
+		// Or can we just ignore this
```
mint-dewit, line 75: You might want to clear the mocks here and reset the timeMock

packages/timeline-state-resolver/src/integrations/websocketTcpClient/__tests__/websocketTcpClient.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+	})
+
+	describe('Connections', () => {
+		test('init', async () => {
+			expect(MockWebSocketTcpConnection.prototype.connect).toHaveBeenCalled()
+		})
+
+		test('terminate', async () => {
+			await device.terminate()
```
mint-dewit, line 84: This could cause race conditions when this test is ran in parallel with other tests

mint-dewit (changes requested):
In general it seems this PR hasn't been linted and formatted. It is recommended to have automatic formatting on save enabled in the Sofie projects, and commit hooks to fail when the linting doesn't pass on commit.

I am quite confused as to why an additional device integration is being added for TCP connections when we already have the TCPSendDevice. There appears to be no benefit to having the TCP connection be an option in this integration.

packages/timeline-state-resolver/src/integrations/websocketTcpClient/$schemas/options.json
mint-dewit: Elsewhere it seems to be implied that the entire `tcp` object should be optional. I am not sure the Sofie UI will handle that well, and also it doesn't seem to be reflected in this JSON

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
olzzon: Maybe I've misunderstood something, but from looking through the code in the other integrations, it does look like the actions payload need to be a Record<string, xxx> type

So I changed the type from any it to:
`payload?: Record<string, SendWebSocketMessagePayload>`

And the iterate over the Record and send the message directly to the connection.

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
olzzon: Sending message directly to connecting instead

packages/timeline-state-resolver/src/integrations/websocketTcpClient/index.ts
mint-dewit: The other integrations are using a loosely typed object which is not ideal but works so long as the implementation is in line with the actions.json file. `payload?: Record<string, SendWebSocketMessagePayload>` however is still not in line with that file nor is your current implementation.

olzzon: has been updated to respect the command structure and use the sendCommand()

nytamin (approved):

PR #377, opened 2025-04-23
feat: make tsr action execution methods be strongly typed

Julusian:
## About the Contributor
This pull request is posted on behalf of Superfly

## Type of Contribution

This is a:  Code improvement

## Current Behavior

The TSR actions for each device have json schemas describing them with some generated types, but the implementation of these in code do not make use of these types very well. This means we are relying on devs to ensure that they get the types correct (spoiler: some are done wrong)

## New Behavior

This reworks the generated types to be a more useful interface, and utilises these in the tsr devices, so that the typings align with typescript being able to check it all.

To do this I opted for making a combined `___DeviceTypes` interface for each integration. This makes the implementations more ergonomic as it allows us to combine all the auto-generated types into a single object instead of needing 3 generic arguments.

At the same time as this, it was trivial to propogate the types for mappings through to the appropriate places, so each device now has the mappings it received typed as only its own which is all it should ever receive.

This fit neatly into the new style of device integrations, and has been mangled into the old style. I haven't put as much effort into the old style.

## Status


mint-dewit:
I really like the general shape this has. I'm left thinking about 2 things:

1. The id parameter when executing the action seems superfluous and I can't intuitively come up with a reason why the actions should be receiving it, perhaps this is a good time to remove it?
2. Some information about using the new `[Name]DeviceTypes` interface could perhaps be added to the [Sofie Docs](https://nrkno.github.io/sofie-core/docs/for-developers/device-integrations/tsr-api)? At the same time I can see that anyone looking at an already implemented integration should be able to grasp it

Julusian:
I've drafted some docs https://github.com/SuperFlyTV/sofie-core/tree/chore/tsr-actions-typings, will open a pr once this one is merged

PR #381, opened 2025-07-01
Bug in commandExecutor: `.preliminary` does not work (SOFIE-4069)

nytamin:
## About the Contributor

This pull request is posted on behalf of the NRK

## Type of Contribution

This is a: Bug fix

## Current Behavior

The CommandExecutor did not execute commands properly if the `.preliminary` property was set.

## New Behavior

(Added unit tests to verify that) CommandExectur now executes commands as expected.

## Testing Instructions

Currently only the Quantel device uses the `preliminary` property, so testing with respect to Quantel playout is prudent.

## Status

… executed properly

### DIFF
--- a/packages/timeline-state-resolver/src/service/commandExecutor.ts
+++ b/packages/timeline-state-resolver/src/service/commandExecutor.ts
@@ -3,8 +3,6 @@ import { BaseDeviceAPI, CommandWithContext } from './device'
 import { Measurement } from './measure'
 import { StateHandlerContext } from './stateHandler'
 
-const wait = async (t: number) => new Promise<void>((r) => setTimeout(() => r(), t))
-
 export class CommandExecutor<DeviceState, Command extends CommandWithContext> {
 	constructor(
 		private logger: StateHandlerContext['logger'],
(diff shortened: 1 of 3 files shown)


packages/timeline-state-resolver/src/service/commandExecutor.ts
```diff
@@ -30,16 +30,27 @@ export class CommandExecutor<DeviceState, Command extends CommandWithContext> {
...
+				const targetTime = start + totalTime - (command.preliminary ?? 0)
+
+				const timeToWait = targetTime - Date.now()
+				if (timeToWait > 0) {
+					await sleep(timeToWait)
+				}
 
 				measurement?.executeCommand(command)
-				return this.sendCommand(command).then(() => {
+				let result: unknown
```
Julusian, line 45: This `result` handling is confusing.

As far as I can tell, this returns `void`, so `result` should always contain `void`. 

Additionally, this flow makes the catch look weird, as it doesnt set `result`, which means that it if was returning something before it has become `| undefined`

packages/timeline-state-resolver/src/service/commandExecutor.ts
```diff
@@ -55,17 +64,31 @@ export class CommandExecutor<DeviceState, Command extends CommandWithContext> {
 
 		await Promise.allSettled(
 			Object.values<Command[]>(commandQueues).map(async (commandsInQueue): Promise<void> => {
-				for (const command of commandsInQueue) {
-					const timeToWait = totalTime - (Date.now() - start)
-					if (timeToWait > 0) await wait(timeToWait)
+				try {
+					for (const command of commandsInQueue) {
+						const targetTime = start + totalTime - (command.preliminary ?? 0)
```
Julusian, line 69: Another thought, it looks like _executeCommandsSalvo and _executeCommandsSequential are now doing the exact same 15 LOC for each command. Perhaps this should be broken out into a new method to avoid the duplication?

Julusian (approved):

packages/timeline-state-resolver/src/service/commandExecutor.ts
yaredTura: Yared Teshoma tura

PR #407, opened 2025-11-13
Feat: Improve json schema

rjmunro:
## About the Contributor

This pull request is posted on behalf of the BBC.

## Type of Contribution

This is a Code improvement

## Current Behavior

When trying to make changes to `packages/timeline-state-resolver-types/src/generated/atem.ts`, I couldn't immediately work out what I had to change to affect the file.

Eventually I found `packages/timeline-state-resolver/src/integrations/atem/$schemas/mappings.json`, but I was getting a warning from VSCode telling me that the `$schema` property itself wasn't allowed in the schema.

## New Behavior

* A bit more info is printed in the headers of generated typescript (although it's possibly not enough as some files seem to be generated from other files that are in turn generated).
* `$schema` is allowed in the schema

## Testing Instructions

Really only comments have changed.

## Status

### DIFF
--- a/packages/timeline-state-resolver-api/$schemas/mapping-schema.json
+++ b/packages/timeline-state-resolver-api/$schemas/mapping-schema.json
@@ -3,6 +3,10 @@
 	"description": "This is a meta-schema that defines the schemas for *device*/$schemas/mappings.json",
 	"type": "object",
 	"properties": {
+		"$schema": {
+			"description": "The $schema property is the schema that this file adheres to",
+			"type": "string"
+		},
 		"default": {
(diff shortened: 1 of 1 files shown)


Julusian (commented):
I'm not sure that this should say `run "yarn generate-schema-types"` because that name will vary depending on where this is used. 

For these schemas inside of the tsr repository that name is correct, but anyone making a tsr plugin will be using the same `tsr-schema-types` tool through whatever script they want. It may or may not be `yarn generate-schema-types` for them, and may not even be `yarn`


But I like including the file path in the banner

PR #430, opened 2026-02-11
fix: rework generated types to minimise variation in type union

Julusian:
## About the Contributor

This pull request is posted on behalf of the BBC

## Type of Contribution

This is a: Bug fix

## Current Behavior

This was a weird one, we appear to be hitting a limit on the size of a type union that appears as soon as one more device type is added:

image

Deleting any of the existing integration types makes the error go away.  
Updating typescript to 5.7 did not resolve it.

Seems to be a designed limitation in typescript: https://github.com/microsoft/TypeScript/issues/40803  
Not 100% sure what this change does that avoids that, but it  avoids the issue

## New Behavior

Instead this reworks the types a little bit, which makes the error go away. Other than a few changes in generics, functionally nothing is really different.

We are not using this Base type in sofie-core, so this should have minimal impact outside of TSR.

## Status

## Problem

TypeScript compilation was hitting a type union complexity limit where adding one more device type triggered a cascade of compilation errors. These errors surfaced in `ConnectionManager.ts` with multiple TS2345 type assignment incompatibilities. Deleting existing device types would eliminate the errors, and upgrading to TypeScript 5.7 did not resolve the issue. The root cause was excessive variation in how `DeviceOptionsBase` type unions were being formed across the codebase.

## Solution

Reworked generated type definitions to minimize variation in type unions by making the generic type parameterization more explicit and consistent throughout the codebase.

### Type Definition Changes (`packages/timeline-state-resolver-types/src/device.ts`)
- Refactored `DeviceOptionsBase` from a single generic parameter (`T`) to two explicit parameters: `TType extends DeviceType` and `TOptions`
- Updated property signatures:
  - `type: DeviceType` → `type: TType`
  - `options?: T` → `options?: TOptions`

### Generated Type Schema (`packages/timeline-state-resolver-tools/bin/schema-types.mjs`)
- Added `Type` field to generated per-directory `DeviceTypes` interfaces
- Converted per-directory `DeviceOptions` declarations from interfaces to type aliases using the new `DeviceOptionsBase<DeviceType.{DeviceTypeId}, {Dir}Options>` form
- Added import of `DeviceType` to generated output files

### Type Propagation Updates
- **`device.ts`**: Updated `Device` and `DeviceWithState` generic constraints to include `Type: DeviceType` in the `DeviceTypes` constraint and pass both `DeviceTypes['Type']` and `DeviceTypes['Options']` to `DeviceOptionsBase`
- **`conductor.ts`**: Updated internal function signatures to use `DeviceOptionsBase<any, any>` consistently across `_mapAllConnections`, device mapping, and filtering operations
- **`deviceContainer.ts`**: Updated `DeviceContainer` and its static `create` method to use the two-parameter `DeviceOptionsBase<DeviceType, unknown>`
- **`ConnectionManager.ts`**: Standardized all public event signatures and method return types to use `DeviceOptionsBase<any, any>`
- **`remoteDeviceInstance.ts`**: Updated `BaseRemoteDeviceIntegration` and `RemoteDeviceInstance` generic constraints to use `DeviceOptionsBase<any, any>`

## Impact

- Functional behavior is unchanged; this is purely a type system restructuring
- The `DeviceOptionsBase` base type changes are isolated within TSR and should have minimal impact on sofie-core, which does not directly depend on this base type
- Resolves the TypeScript compilation errors in `ConnectionManager.ts` by reducing union type complexity

### DIFF
--- a/packages/timeline-state-resolver-tools/bin/schema-types.mjs
+++ b/packages/timeline-state-resolver-tools/bin/schema-types.mjs
@@ -227,7 +227,7 @@ for (const dir of dirs) {
 
 	const dirId = capitalise(dir)
 
-	let output = ''
+	let output = "import type { DeviceType } from './device-options.js'\n"
 
 	// compile options from file
 	try {
@@ -401,14 +401,6 @@ ${actionDefinitions
 			}"\n` + output
 	}
 
-	output += `
-export interface ${dirId}DeviceTypes {
-	Options: ${dirId}Options
-	Mappings: SomeMapping${dirId}
-	Actions: ${actionDefinitions.length > 0 ? `${dirId}ActionMethods` : 'null'}
-}
-`
-
 	let deviceTypeId = toConstantCase(dir)
 	// Special case handling for some devices, for backwards compatibility
 	if (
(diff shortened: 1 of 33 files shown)


nytamin (approved):
I can confirm this solves an issue for me as well!

PR #440, opened 2026-02-24
fix: add extendable DeviceOptionsMap type for tsr-plugins

Julusian:
## About the Contributor

This pull request is posted on behalf of Superfly

## Type of Contribution

This is a: Bug fix / Feature 

## Current Behavior

When using a tsr-plugin it is possible to extend the timeline content types, but not the device options types.

## New Behavior

This gives the device options similar treatment

TODO: The docs in sofie-core should be updated with an example, and to note that this `tsr-extend.d.ts` should be referenced in the tsconfig `types` array:
```
{
	"compilerOptions": {
		"types": ["node", "./src/tsr-extend.d.ts"],
	}
}
```

## Status


scriptorian (approved):
All tested and working well

